PR #64, opened 2026-06-26
fix: warm cache on client cancel to break cold-owner loop

### LINKED ISSUE #63: Cold-owner cancellation loop: cache never warms because populate is gated on the client request surviving
## Summary

When one TAG node is **cold** (empty local cache) but still owns its share of the consistent-hash ring, it pins at **40–70 CPU cores** indefinitely while warm peers sit at 8–12, and its cache never warms. Reported by Goldsky (Prism team), `mercury-prod`, on `tigrisdata/tag:v1.9.2`.

**Root cause:** cache population is gated on the *client request surviving*. For a cold owner, the cold fetch from Tigris is slower than the client/peer deadline, so the client cancels mid-fetch (`context canceled`); the in-flight fetch is abandoned and nothing is cached, so the **same** large compacted SSTs are re-fetched on every request — a self-sustaining cold-miss/cancel loop with no forward progress.

This is operationally important because the standard recovery for the *separate* nil-deref crash defect is to **wipe a node's cache** (PVC delete + cold restart), which produces exactly this cold-owner precondition. So the crash remediation reliably triggers this loop.

> Note: the customer report attributes this to "ocache abandoning the upstream fetch." That is correct about the symptom but wrong about the layer — **ocache has no upstream-fetch logic** (it only stores what is explicitly `Put` into it, and the cache `Put` TAG issues is already on a detached context). The bug is entirely in TAG's request orchestration. **No ocache change is required.**

### A — full-object miss (`proxy/get_object.go`)

`fetchAndBroadcast` runs the upstream fetch on the **client request context**:

- `streamFromUpstream(ctx, …)` — `get_object.go:245`
- upstream GET issued on that ctx — `get_object.go:269` (`DoRequestWithCreds(ctx, …)`)
- the read/broadcast loop returns on `ctx.Done()` — `get_object.go:323-326`

The cache *writer* is correctly detached (`caching.go:129`, `context.WithTimeout(context.Background(), …)`), but it is **fed** by `streamFromUpstream`. On client cancel the source dies, `broadcaster.Complete(context.Canceled)` starves the cache listener, and the write never finalizes. Detaching only the writer is insufficient — the **source** must be detached too.

### B — range miss (the hot path for this workload): `handleRangeWithBackgroundCache`

The customer's objects are read as **byte ranges**, so this is the dominant path. The background full-object fetch that warms the cache **is** already detached and deduplicated (`triggerBackgroundCacheFetch` → `context.Background()` + `activeBackgroundFetches` `LoadOrStore`, `caching.go:371-407`). The bug is purely **when** it is triggered:

```go
// proxy/get_object.go  (handleRangeWithBackgroundCache)
n, err := io.Copy(w, resp.Body)          // :618  stream range to client (client ctx)
if err != nil {
    return err                            // :620-623  client cancel -> early return
}
...
// :627  "Trigger background fetch AFTER streaming completes"
if resp.StatusCode == 206 && … {
    s.triggerBackgroundCacheFetch(…)      // :641  NEVER REACHED on cancel
}
```

When the client deadline is shorter than the cold fetch, `io.Copy` fails mid-stream, the function returns at `:622`, and the warming fetch at `:641` never fires. Next range request -> same miss -> same cancel -> the loop.

## Why it presents as observed

- Cold owner does fetch+decompress for 100% of its slice and caches none of it -> 40–70 cores.
- Warm peers serve from disk but originate the gRPC-forward cancellations toward the cold owner -> cheap CPU, highest cancel counts.
- `Failed to forward request` / `context canceled` = TAG's upstream GET aborting (`proxy/forwarder.go`); `Failed to stream range from cache` = the inter-node gRPC range read aborting on the same client ctx.
- Even between OOM restarts the node warms far slower than a clean refill, because cancellations keep discarding completed work.

### B — load-bearing fix for this customer

Move `triggerBackgroundCacheFetch` so it fires after `io.Copy` **regardless of outcome** (success or client cancel) — e.g. via a `defer` set up once the response is known cacheable (206 + total size within threshold + creds + cache enabled). It is already detached + deduplicated and guarded by `GetMeta(...!found)`, so this is safe, preserves the happy-path "after streaming" ordering (the defer still runs after `io.Copy`), and warms the key even when the client walks away. The dedup + `!found` guard prevent redundant fetches.

### A — full-object detach

Run `streamFromUpstream` on `context.WithTimeout(context.WithoutCancel(ctx), <upstreamTimeout>)` rather than the raw client ctx, so the fetch+broadcast (which feeds the already-detached cache writer) completes even after the client cancels. Keep `writeChunksToResponse` on the client ctx so the client side still cancels normally. (`WithoutCancel` preserves request-scoped values like trace IDs but drops cancellation/deadline, hence the explicit timeout.)

### Optional — observability (maps to report ask B3)

TAG already emits background-fetch triggered/succeeded/failed counters. Add an explicit **cancel-before-populate** counter (count + bytes) on path A so a cold node failing to warm is visible as a metric, not just a log storm + CPU.

## Expected behavior after fix

The first fetch for a key populates the cache (subject to TTL/size policy) even if the triggering request was canceled, so the second request hits and the loop terminates after one fetch per key.

## Reproduction

1. Bring up (or wipe) one node so it is cold while it still owns a share of the ring.
2. Drive byte-range read traffic for its keyspace where each object is a multi-MB compacted SST on Tigris and the client/peer deadline is shorter than a cold fetch.
3. Observe continuous `context canceled` on the upstream `Get`, the same keys re-fetched repeatedly, the cache not warming, and CPU pinned.

## Environment

- Image `tigrisdata/tag:v1.9.2` (build `d1656a8`, 2026-04-17); ocache `storage/server/coordinator @v1.5.1`; Go 1.24.2/arm64; grpc-go v1.72.2.
- Cluster mode, 3-node StatefulSet, consistent-hash ring; client HTTP `:8080`, inter-node gRPC `:9000`, gossip `:7000`/UDP; upstream `https://t3.storage.dev`; `ttl=720h`, `size_threshold=1GiB`, `max_disk_usage≈2Ti`, 40Gi container memory limit.
- Example canceled keys: `prism-arbitrum-index/blocks/v1/tx_hash/compacted/01KVYR02QKY4RC48FD320J09Z9.sst`, `prism-base-index/blocks/v2/bloom/compacted/01KT93YAJ4X53HAVB70T85SF7Q.sst`.

## Observed rates (mercury-prod, 2026-06-25 ~08:10–08:26 UTC)

- Cold-owner CPU 40–70 cores sustained (warm peers 8–12).
- `Failed to forward request`: ~69k / 15 min fleet-wide (~77/s).
- `Failed to stream range from cache`: per pod ≈ 31.2k / 30.8k / 10.0k per 15 min.
- All pods pinned ~39–39.5 GiB against the 40 GiB limit; cold owner OOM-resets to ~0 and climbs again.

ovaistariq:
## Summary

Fixes #63.

A **cold** TAG node that still owns its share of the consistent-hash ring pins CPU at 40–70 cores and never warms its cache because cache population was gated on the **client request surviving**. When the client/peer deadline is shorter than a cold fetch from Tigris, the client cancels mid-fetch (`context canceled`), the in-flight fetch is abandoned, nothing is cached, and the same large compacted SSTs are re-fetched on every request — a self-sustaining cold-miss/cancel loop with no forward progress. This is the precondition produced by the standard "wipe the PVC + cold restart" remediation, so it triggers reliably.

Both instances of the root cause live in `proxy/get_object.go`. No ocache change is required (the cache `Put` is already on a detached context — the bug is in TAG's request orchestration).

### B — range path (the hot path for this workload): `handleRangeWithBackgroundCache`
Trigger the background full-object fetch via `defer` so it fires after `io.Copy` **regardless of outcome** (success or client cancel). The fetch is already detached (`context.Background()`) + deduplicated (`activeBackgroundFetches`) and guarded by `GetMeta(...!found)`, so warming the key on cancel is safe and never issues redundant fetches. Happy-path "after streaming" ordering is preserved (the defer still runs after `io.Copy`).

### A — full-object path: `fetchAndBroadcast`
Run `streamFromUpstream` on `context.WithTimeout(context.WithoutCancel(ctx), upstreamFetchTimeout)` so the fetch that feeds the already-detached cache writer completes even after the client cancels. `writeChunksToResponse` stays on the client ctx so the client side still cancels normally. `WithoutCancel` preserves request-scoped values (trace IDs) while dropping cancellation/deadline; the explicit timeout (5m, mirroring the upstream HTTP client timeout) bounds the now-uncancelable fetch.

## Expected behavior after fix

The first fetch for a key populates the cache (subject to TTL/size policy) even when the triggering request is canceled, so the second request hits and the loop terminates after one fetch per key.

## Tests

- New `NewTestEnvironmentWithCacheHandler` — a cache-enabled integration env backed by a custom upstream handler, for precise control over upstream responses (e.g. holding/cutting a stream mid-flight).
- `TestColdOwner_RangeWarmsCacheOnClientCancel` (path B) and `TestColdOwner_FullObjectWarmsCacheOnClientCancel` (path A): cancel a request mid-stream and assert the full object still lands in cache.
- Full unit suite, cache integration suite, `go vet`, and `gofmt` all pass.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path GetObject orchestration for cache misses; mitigations include deduplication, GetMeta guards, and skipping background warm on genuine upstream failures.
> 
> **Overview**
> Fixes **#63** by stopping the cold-miss/cancel loop where a short client deadline aborts the inline fetch before anything is cached, so the same keys are re-fetched forever on a cold TAG node.
> 
> On **full-object** cache misses (`fetchAndBroadcast`), after the client context is gone it waits for the upstream fetch outcome and, if the object is still uncached and the failure was cancellation/deadline (not a real upstream error), kicks off the existing **deduplicated** `triggerBackgroundCacheFetch` instead of leaving the key cold.
> 
> On **range** cache misses (`handleRangeWithBackgroundCache`), the background full-object warm is scheduled with a **`defer`** so it runs after `io.Copy` whether streaming succeeds or fails (client disconnect / mid-stream error), still guarded by cacheability checks and `GetMeta(!found)`.
> 
> Adds **`NewTestEnvironmentWithCacheHandler`** for integration tests with a controllable upstream, plus unit tests for warm vs no-warm on upstream failure, and integration tests that cancel mid-stream and assert the full object lands in cache.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bb932b8aac0bd243590f316369b3023da0254b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/cold_owner_test.go
+++ b/proxy/cold_owner_test.go
@@ -1,0 +1,85 @@
+package proxy
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/tigrisdata/tag/proxy/broadcast"
+)
+
+// warmForwarder drives fetchAndBroadcast's cold-owner warm branch in isolation.
+// The inline upstream fetch (DoRequestWithCreds) returns doRequestErr;
+// DoFullObjectRequest — reached only via the deduplicated background warm —
+// reports the bucket/key it was asked to fetch on fullObjectFetch.
+type warmForwarder struct {
+	mockForwarder
+	doRequestErr    error
+	fullObjectFetch chan [2]string
+}
+
+func (m *warmForwarder) DoRequestWithCreds(_ context.Context, _ *http.Request, _, _ string) (*http.Response, error) {
+	return nil, m.doRequestErr
+}
+
+func (m *warmForwarder) DoFullObjectRequest(_ context.Context, bucket, key, _, _ string) (*http.Response, error) {
+	m.fullObjectFetch <- [2]string{bucket, key}
+	return nil, errors.New("warm: stop background fetch")
+}
+
+// runFetchAndBroadcast drives fetchAndBroadcast with a pre-canceled context (so
+// writeChunksToResponse returns promptly) and the given inline-fetch error.
+func runFetchAndBroadcast(t *testing.T, doRequestErr error) *warmForwarder {
+	t.Helper()
+	mock := &warmForwarder{
+		doRequestErr:    doRequestErr,
+		fullObjectFetch: make(chan [2]string, 1),
+	}
+	svc, _ := newTestService(mock, true) // empty in-memory cache → GetMeta reports !found
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // the cold-owner peer/client deadline has already expired
+
+	bcaster := broadcast.NewBroadcaster(broadcast.DefaultChannelBuffer)
+	req := httptest.NewRequest(http.MethodGet, "/cold-bucket/cold-key", nil)
+	w := httptest.NewRecorder()
+
+	_ = svc.fetchAndBroadcast(ctx, w, req, "cold-bucket", "cold-key", "access", "secret", bcaster, time.Now(), "MISS")
+	return mock
+}
+
+// TestFetchAndBroadcast_WarmsOnContextCancel verifies that when the client
+// context is canceled before the inline fetch populates the cache (issue #63
+// path A) and the upstream was healthy (the fetch was aborted by cancellation,
+// surfacing a context error), fetchAndBroadcast hands warming off to the
+// deduplicated background fetcher.
+func TestFetchAndBroadcast_WarmsOnContextCancel(t *testing.T) {
+	mock := runFetchAndBroadcast(t, context.Canceled)
+
+	select {
+	case got := <-mock.fullObjectFetch:
+		if got[0] != "cold-bucket" || got[1] != "cold-key" {
+			t.Fatalf("background warm fetched %q/%q, want cold-bucket/cold-key", got[0], got[1])
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatal("expected a background warm fetch after a canceled request, but none occurred")
+	}
+}
+
+// TestFetchAndBroadcast_NoWarmOnUpstreamError verifies that when the upstream
+// itself failed (not the client), fetchAndBroadcast does NOT warm — even though
+// the client context is canceled — so a doomed retry never amplifies load
+// against an already-unhealthy upstream.
+func TestFetchAndBroadcast_NoWarmOnUpstreamError(t *testing.T) {
+	mock := runFetchAndBroadcast(t, errors.New("connection refused"))
+
+	select {
+	case got := <-mock.fullObjectFetch:
+		t.Fatalf("unexpected background warm fetch %q/%q on an upstream failure", got[0], got[1])
+	case <-time.After(200 * time.Millisecond):
+		// No warm — correct.
+	}
+}
--- a/proxy/get_object.go
+++ b/proxy/get_object.go
@@ -33,14 +33,6 @@ var bufferPool = sync.Pool{
 // allocation, and synchronization overhead for small objects.
 const smallObjectThreshold = 64 * 1024 // 64KB
 
-// upstreamFetchTimeout bounds the detached upstream fetch in fetchAndBroadcast.
-// The fetch runs on a context derived from the client request but with
-// cancellation/deadline stripped (context.WithoutCancel), so a client cancel
-// no longer abandons the in-flight fetch that feeds the cache writer. This
-// timeout caps how long that detached fetch may run. It mirrors the upstream
-// HTTP client timeout (see forwarder.go) so the two bounds stay aligned.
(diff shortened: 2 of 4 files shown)


efirs (approved):

ovaistariq:
@greptile

ovaistariq:
@greptile

PR #72, opened 2026-07-01
fix: bound S3 ingress + cache-populate concurrency (thread-exhaustion/OOM fix)

ovaistariq:
## Problem

Under load — especially with a degraded cache ring — TAG's S3 ingress had **no working concurrency limit**. Every request spawned handlers that block on disk reads and cgo RocksDB calls (each pinning an OS thread), and cache populates spawned unbounded streaming-write pipelines, until the process hit Go's 10k-thread cap (`fatal error: thread exhaustion`) or OOM'd. The one existing limiter, `proxy.Service.cacheSemaphore`, was created but **never acquired** — it bounded nothing.

This is the TAG-side keystone for ocache issue #162 (thread exhaustion / admission control). TAG embeds ocache in-process (`embedded.New`), so its request load never traverses ocache's gRPC interceptors — admission must live here, at the S3 edge.

## Changes

**1. Global ingress admission middleware** (`handlers`) — a counting semaphore bounds concurrently-served S3 requests; when full, sheds with **`503 SlowDown`** so overload becomes backpressure instead of unbounded goroutine/thread/memory growth. Only exact `/health` and `/metrics` (and `/debug/pprof/` when pprof is enabled) are exempt, so S3 objects in a bucket named `health`/`metrics`/`debug` cannot bypass admission. Configurable via `server.max_inflight_requests` (0/unset = default 1024; negative disables).

**2. Wire the dead cache-populate limiter** (`proxy`) — `cacheSemaphore` is now acquired in `setupCacheListener`. When the concurrent-cache-write limit is saturated, TAG **skips caching** (still serves/forwards from upstream) rather than spawn unbounded populate work. Configurable via `cache.max_concurrent_writes` (0/unset = default 100; negative disables).

**3. Remove the whole-object-buffering footgun** (`cache`) — audited the GET path for 256MB+ (SlateDB SST) safety. It already streams: hits via `GetBodyStream` (buffering only objects `<= 64KB`, streaming the rest via `io.Pipe`), ranges via `GetRangeStream`, populate via `PutWithMetaStreamTombstoneAware`. Removed the unused `GetWithMeta`, which read whole bodies into a `bytes.Buffer` (a latent OOM path for large objects).

New metrics: `tag_inflight_requests`, `tag_admission_shed_total`, `tag_cache_populate_skipped_total`. Unit tests for admission (shed/exempt/bucket-not-exempt/unlimited) and the cache-slot limiter. Full `make test` passes.

## Out of scope (ocache-side, separate PRs)
Bounding the internal peer-forward fan-out and setting `grpc.MaxConcurrentStreams` on ocache's inter-node surface (issue #162 Part B), which requires an ocache release TAG then bumps to.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path behavior under overload (503 shedding and degraded cache fill) and raises default concurrent cache writes from 100 to 256; operational endpoints remain reachable but mis-tuned limits could increase client-visible SlowDown or reduce cache warmth.
> 
> **Overview**
> Adds **overload protection** at the S3 edge and on the cache-populate path so load turns into **503 SlowDown** and skipped writes instead of unbounded goroutines, threads, and memory.
> 
> **Ingress admission** (`handlers`): middleware limits concurrently served S3 requests via `server.max_inflight_requests` (default 1024; negative disables). Saturated requests get S3 **SlowDown**; `/health`, `/metrics`, and registered `/debug/pprof/*` routes stay exempt using **mux path templates** so S3 keys like `/debug/pprof/...` cannot bypass the limit. Metrics: `tag_inflight_requests`, `tag_admission_shed_total`.
> 
> **Cache-write limiter** (`proxy`): `cache.max_concurrent_writes` (default **256**, was a fixed 100 slot that was never acquired) now gates `setupCacheListener` and background `fetchFullObjectToCache`; when full, responses still stream from upstream but caching is skipped (`tag_cache_populate_skipped_total`). Background fetches reserve a slot before upstream and treat saturation as `errCachePopulateDeclined` (not a fetch failure).
> 
> **Config/docs**: YAML and env vars `TAG_MAX_INFLIGHT_REQUESTS`, `TAG_CACHE_MAX_CONCURRENT_WRITES`; `NewServer` takes `maxInflight`.
> 
> **Cleanup**: removes unused `cache.GetWithMeta` (whole-body buffer read) and related tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dde1385ad72e41a7d7293a5dea2a112d4e5ea371. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -65,7 +65,7 @@ This covers all private repos under the `tigrisdata` org (e.g., `ocache`). Witho
 - **Two-Key Cache Storage**: Objects stored as `meta|bucket|key` (headers/ETag) and `body|bucket|key` (raw bytes)
 - **Tombstone Invalidation**: Writes tombstone marker before DELETE to prevent stale async cache writes
 - **Ingress Admission**: `server.max_inflight_requests` (default 1024) bounds concurrently-served S3 requests; excess is shed with 503 SlowDown. Operational endpoints (`/health`, `/metrics`, `/debug/pprof/*`) are exempt. `0`/unset = default, negative = disabled.
-- **Semaphore-Gated Cache Writes**: `cache.max_concurrent_writes` (default 100) bounds concurrent cache-populate operations; when saturated, objects are served from upstream without being cached. `0`/unset = default, negative = disabled.
+- **Semaphore-Gated Cache Writes**: `cache.max_concurrent_writes` (default 256) bounds concurrent cache-populate operations; when saturated, objects are served from upstream without being cached. `0`/unset = default, negative = disabled.
 
 ## Key Dependencies
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,7 +64,8 @@ This covers all private repos under the `tigrisdata` org (e.g., `ocache`). Witho
 - **Broadcast/Subscriber Pattern**: Request coalescing via `proxy/broadcast/` - streams chunks to multiple listeners simultaneously
 - **Two-Key Cache Storage**: Objects stored as `meta|bucket|key` (headers/ETag) and `body|bucket|key` (raw bytes)
 - **Tombstone Invalidation**: Writes tombstone marker before DELETE to prevent stale async cache writes
-- **Semaphore-Gated Background Ops**: `cacheSemaphore` (100 concurrent) limits background cache writes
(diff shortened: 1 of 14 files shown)


ovaistariq:
@greptile

ovaistariq:
@greptile

ovaistariq:
@greptile

ovaistariq:
@greptile

ovaistariq:
@greptile

PR #78, opened 2026-07-09
feat: publish native install assets to the release bucket

ovaistariq:
## Context

Native installs fetched `config.yaml` from `raw.githubusercontent.com`, which coupled installs to GitHub (rate limits, only works while the repo is public) and pulled config from a ref *separate* from the versioned binary. The binaries already live in the `tag-releases` bucket (`t3.storage.dev`), uploaded per-version and to `latest/` by the `Upload Release to T3` workflow. This publishes the native assets to that same bucket so installs are reproducible and GitHub-free.

## Changes

- **`.github/workflows/t3-upload-release.yaml`** — after the binary upload, take `config.yaml`, `install.sh`, `run.sh` **from the exact release tag** (so they match the binary) and upload them to `s3://tag-releases/<version>/`, plus `s3://tag-releases/latest/` on release. Each uploaded script's default `TAG_VERSION` is rewritten to the release version (or `latest` for the `latest/` copies), so `curl .../<dir>/install.sh | bash` installs the matching version. Releases predating `deploy/native/` are detected and skipped.
- **`deploy/native/install.sh`** — fetch config from `${TAG_RELEASES_URL}/${TAG_VERSION}/config.yaml` (same base URL and version as the binary) instead of `raw.githubusercontent.com`. On a miss, warn and note that TAG can be configured entirely via env vars. This retires the GitHub-raw config path (and the tag-ref/`main` fallback added earlier).
- **`README.md`** — document the one-liner bootstrap:
  ```bash
  curl -fsSL https://tag-releases.t3.storage.dev/latest/install.sh | bash
  curl -fsSL https://tag-releases.t3.storage.dev/v1.10.0/install.sh | bash
  ```

## Resulting bucket layout

```
s3://tag-releases/<version>/   tag-linux-amd64  tag-linux-arm64  tag-darwin-arm64
                               config.yaml  install.sh  run.sh   <-- new
s3://tag-releases/latest/      (same set, scripts default to 'latest')
```

## Verification

- ✅ `bash -n deploy/native/install.sh`.
- ✅ `sed` version-rewrite simulated against the real scripts — only the `TAG_VERSION` definition line changes; version-pinned and `latest` flavors both correct.
- ✅ Workflow YAML parses (yq, 11 steps).

## Notes

- The bucket-published `install.sh` for a version always has a sibling `config.yaml` in the same dir, so the bootstrap path has no version-skew.
[...]

PR #80, opened 2026-07-09
fix: release-PR follow-ups for native install and k8s cache cluster

ovaistariq:
Follow-up fixes for review comments on #78 (native install assets in the release bucket), kept as a separate PR.

## Changes

- **`deploy/native/run.sh` — stale `latest` binary cache.** With `TAG_VERSION=latest` the cache path is `.bin/tag-latest`; after one download, later starts short-circuited on the cached file even after `latest/tag-*` was updated, pinning users to an old build. Now `latest` re-downloads every start; pinned versions still cache.
- **`deploy/native/run.sh` — installed config ignored.** `install.sh` writes `/etc/tag/config.yaml`, but the runner started `tag` without `--config`, so edits for TLS/upstream/cache were silently dropped in favor of defaults + env. The runner now passes `--config ${TAG_CONFIG_FILE}` (default `/etc/tag/config.yaml`, set empty to opt out) when the file exists. Empty-array expansion is bash-3.2-safe for macOS.
- **`deploy/kubernetes/base/statefulset.yaml` — headless service as seed.** `TAG_CACHE_SEED_NODES` was the headless service name, which resolves to a rotating set of pod IPs and gives gossip no deterministic peer list (pods could form split/partial clusters). Now seeds from stable per-pod DNS (`tag-0/1/2.tag-headless:7000`); scaled-up replicas join via these seeds.
- **`.github/workflows/t3-upload-release.yaml` — manual version skips assets.** Manual dispatch used the entered version verbatim as both the S3 directory and the git ref for native assets; entering `1.10.0` for tag `v1.10.0` left `install.sh`/`run.sh` missing from the release dir. The version is now normalized to the `v`-prefixed tag so the S3 path, `gh release` lookups, and native-asset git ref all agree.

## Verification

- ✅ `bash -n deploy/native/run.sh`; empty-array expansion validated under `set -u` (no-args → nothing, with-args → `--config <path>`).
- ✅ Workflow YAML parses (yq).
- ✅ `kubectl kustomize deploy/kubernetes/base` renders the stable seed list.

## Note (out of scope, flagged)

The runner's TLS-aware health check keys off `TAG_TLS_CERT_FILE`/`TAG_TLS_KEY_FILE` env vars (the flow documented in `docs/tls.md`). Now that the runner also honors the config file, enabling TLS *purely via config* (non-default; the shipped config leaves cert/key empty) would make health probes use `http` against an `https` listener. Happy to follow up by having the runner also detect TLS from the config file if you want that covered.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to shell runner logic and CI version normalization; no application auth, data paths, or core service code.
> 
> **Overview**
> Hardens **native install/run** and **manual T3 release uploads** so version tags, binaries, and `/etc/tag/config.yaml` stay aligned in production.
> 
> **`deploy/native/run.sh`** now passes **`--config`** when `TAG_CONFIG_FILE` exists (default `/etc/tag/config.yaml`), so installer-written TLS/upstream/cache settings apply instead of being ignored. Health checks use **HTTPS** when TLS is set via env **or** non-empty `tls_cert_file` / `tls_key_file` in that file. **`TAG_VERSION=latest`** re-downloads every start; pinned versions still cache. Downloads go through a **temp file + rename** so a failed curl cannot wipe a working cached binary.
> 
> **`.github/workflows/t3-upload-release.yaml`** normalizes manual dispatch input (e.g. `1.10.0` → `v1.10.0`) so S3 paths, `gh release` lookups, and native asset git refs match real release tags.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b16ea549b59c59fcb53acbd0edf9426e25b424a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/t3-upload-release.yaml
+++ b/.github/workflows/t3-upload-release.yaml
@@ -37,8 +37,15 @@ jobs:
           WORKFLOW_RUN_SHA: ${{ github.event.workflow_run.head_sha }}
         run: |
           if [ -n "$INPUT_VERSION" ]; then
-            echo "version=$INPUT_VERSION" >> $GITHUB_OUTPUT
-            echo "Using manual version: $INPUT_VERSION"
+            # Normalize to the v-prefixed tag so the S3 directory, gh release
+            # lookups, and the native-asset git ref all reference the same tag
+            # (release tags are always vX.Y.Z). Entering "1.10.0" -> "v1.10.0".
+            case "$INPUT_VERSION" in
+              v*) VERSION="$INPUT_VERSION" ;;
+              *)  VERSION="v$INPUT_VERSION" ;;
+            esac
+            echo "version=$VERSION" >> $GITHUB_OUTPUT
+            echo "Using manual version: $VERSION"
           else
             # Get the tag pointing to the workflow_run commit SHA
             git fetch --tags
--- a/deploy/kubernetes/base/statefulset.yaml
+++ b/deploy/kubernetes/base/statefulset.yaml
@@ -55,8 +55,12 @@ spec:
               value: ":9000"
             - name: TAG_CACHE_ADVERTISE_ADDR
               value: "$(POD_NAME).tag-headless:9000"
+            # Seed from stable per-pod DNS names (StatefulSet ordinals 0-2 via the
+            # headless service), not the headless service name itself -- the latter
+            # resolves to a rotating set of pod IPs and gives gossip no deterministic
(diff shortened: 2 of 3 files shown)


ovaistariq:
@greptile

singler (approved):

Issue #92, opened 2026-07-14
Frequently updated small file returns incorrect content length and sometimes empty body


pboros:
I have a workload backed by TAG where it has lots of large files that are written once and a few very small metadata file that are updated frequently. 

On requesting the small file and writing it concurrently I am getting `request or response body errors` (bodies are truncated). 

The workload I am using is parseable, tag is a cache, and it's caching parseable's mutable metadata (manifest.json, .schema, ingestor state) that parseable rewrites constantly and reads with read-after-write expectations. Tag serves empty/partial/truncated bodies for those hot keys (its own cache body unavailable: EOF logs on the exact same files), which breaks both queries and staging uploads. If I run TAG in a transparent proxy mode with `TAG_CACHE_DISABLED=true` the issue is gone, but the performance is the non-tag performance. 

I can imagine 2 ways to fix this, one is fixing the root cause to get read after write semantics, the other is to have an allowlist/blocklist so the user is able to bypass caching just for those specific patterns. In my case, if these files were not cached at all, it would be fine, the improvements are coming from caching the parquet files.

ovaistariq:
## Root cause analysis

The truncated/empty bodies are a cache consistency race, not data corruption. TAG stores every cached object as **two independent cache entries** — `meta|bucket|key` (headers, ETag, Content-Length) and `body|bucket|key` (raw bytes) — with no atomicity or version linkage between them, and reads are two separate operations (`GetMeta()` then `GetBodyStream()`). The body key is fixed (no ETag/generation suffix), so every rewrite overwrites the same slot in place.

For a hot key that is rewritten constantly (parseable's `manifest.json`, `.schema`, ingestor state), a reader can pair **version-N metadata with version-N+1 (or deleted) body**. Three concrete races:

1. **Truncation / wrong Content-Length** — `serveFromCache` buffered the small body and emitted `Content-Length` from the (stale) metadata without checking it against the actual buffered body length. Reader loads `meta_v1` (length 100), a concurrent rewrite lands `body_v2` (50 bytes) before the body read → response declares 100 bytes but carries 50 → the `request or response body errors` you're seeing.
2. **Empty body / `cache body unavailable: EOF`** — invalidation deletes meta then body, and PUT/DELETE handlers invalidate the pair while readers are mid-flight. A reader that got meta just before the delete finds the body gone → the exact `cache body unavailable: EOF` log on those hot keys.
[...]

PR #94, opened 2026-07-15
fix: reject meta/body length mismatch when serving small objects from cache

ovaistariq:
Fixes the client-visible truncation / wrong-`Content-Length` / empty-body symptom in #92.

## Root cause (full analysis in the issue)

TAG stores each cached object as two independent entries — `meta|bucket|key` and `body|bucket|key` — with no atomicity or version linkage, and reads them separately (`GetMeta()` then `GetBodyStream()`). For a hot key that is rewritten constantly (Parseable's `manifest.json`, `.schema`, ingestor state), a reader can pair **version-N metadata with version-N+1 (or deleted) body**:

1. **Truncation / wrong length** — `serveFromCache` buffered the small body and emitted `Content-Length` from the (stale) metadata without comparing it to the actual buffered body → response declares 100 bytes but carries 50 (the `request or response body errors`).
2. **Empty body / `cache body unavailable: EOF`** — a reader that got meta just before invalidation deletes the body finds it gone.

## The fix

In the small-object path, `serveFromCache` now serves from cache **only when the buffered body length matches `meta.ContentLength`**. Any mismatch (or unavailable body) is treated as a cache miss; both call sites (cache-hit GET and 304 revalidation) fall through to upstream, and the miss path re-caches — so a torn meta/body pair self-heals.

- `proxy/revalidation.go` — length-match guard + clearer mismatch error.
- `proxy/revalidation_test.go` — unit test for mismatch (nothing written to the response) and an end-to-end 304-revalidation test with a concurrently overwritten body key.

## Scope / caveat

This eliminates races 1 and 2 (the reported symptom). It does **not** fully restore read-after-write: race 3 (a GET racing an in-flight PUT can refetch and re-cache the pre-PUT object) remains, and is addressed by the planned follow-up — **ETag-versioned body keys** + **post-PUT re-invalidation** (see the issue). An optional `cache.skip_patterns` operator escape hatch is also proposed there.

## Verification
- ✅ `make test` (all unit packages) passes, including the two new regression tests.
- ✅ `make test-race` clean across all packages (this is a concurrency fix).

Addresses #92 (symptom fix). The issue stays open for the root-cause follow-up (ETag-versioned body keys + post-PUT re-invalidation).

---

> [!NOTE]
> **Medium Risk**
> Changes core GET/revalidation and stale-serving paths for cached objects; intentional fail-closed behavior during outages when meta/body are inconsistent.
> 
> **Overview**
> Fixes client-visible **wrong `Content-Length`**, truncation, and empty-body errors when cached metadata and body come from different versions of the same hot key (#92).
> 
> For **small objects**, `serveFromCache` now serves from cache only when the buffered body length equals `meta.ContentLength`. Any read error, empty body, or length mismatch is treated like a cache miss: callers can fall through to upstream (e.g. after a **304** revalidation) without committing a bad response.
> 
> **Stale fallback** during upstream failure is **fail-closed** for inconsistent pairs: `serveStaleFromCache` propagates the error and logs a warning instead of serving a skewed body, so clients may see **5xx** in the rare torn-pair-during-outage case rather than corrupt data. Large-object streaming is unchanged.
> 
> Regression tests cover mismatch detection (no response written), stale fail-closed, and **304 → upstream `Forward`** when meta and body disagree.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a54fd609bad78fb13e199f55c64982dc864aa88b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/revalidation.go
+++ b/proxy/revalidation.go
@@ -239,7 +239,13 @@ func (s *Service) serveFromCache(
 		bodyBuf.Reset()
 
 		bodyErr := s.cache.GetBodyStream(ctx, bucket, key, bodyBuf)
-		if bodyErr == nil && bodyBuf.Len() > 0 {
+		// Serve only if the body length matches the metadata's Content-Length.
+		// Meta and body are separate cache entries with no atomicity between
+		// them, so a concurrent overwrite/delete of a hot key can pair one
+		// version's metadata with another version's body. Serving that pair
+		// emits a wrong Content-Length (truncated or over-read response), so
+		// treat any mismatch as a cache miss and fall through to upstream.
+		if bodyErr == nil && int64(bodyBuf.Len()) == meta.ContentLength {
 			metrics.RecordCacheHit()
 			meta.WriteHeaders(w)
 			w.Header().Set(XCacheHeader, XCacheHit)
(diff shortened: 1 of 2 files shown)


pboros (dismissed):

ovaistariq:
Closing in favor of the root-cause fix. Per discussion on the design: ETag-versioned body keys (`body|bucket|key|<etag>`) make the served meta/body pair consistent by construction, which eliminates the torn-pair race entirely (races 1 & 2) for both small and large objects — making this length-guard mitigation redundant. Going straight to the root fix (ETag-versioned bodies + post-PUT re-invalidation) instead of shipping a stopgap. Tracked in #92.

PR #95, opened 2026-07-15
fix: ETag-versioned cache body keys for read-after-write consistency

### LINKED ISSUE #92: Frequently updated small file returns incorrect content length and sometimes empty body

ovaistariq:
Root-cause fix for #92 (truncated/empty small-file bodies under concurrent read+write). Supersedes the closed mitigation PR #94.

## Problem

Each cached object is two independent entries — `meta|bucket|key` and `body|bucket|key` — with no version linkage, and the body slot is fixed (every rewrite overwrites it in place). For a constantly-rewritten hot key (Parseable's `manifest.json`, `.schema`, ingestor state), a reader can pair one version's metadata with another version's body:

1. **Truncation / wrong `Content-Length`** — meta_v1 (len 100) + body_v2 (50 bytes) → `request or response body errors`.
2. **Empty body / `cache body unavailable: EOF`** — a reader that got meta just before invalidation finds the body gone.
3. **Stale re-population (read-after-write)** — a GET racing an in-flight PUT re-caches the pre-PUT object.

## Fix

- **ETag-versioned body keys** — bodies live at `body|bucket|key|<etag>`; a reader derives the body address from `meta.ETag`, so a served meta always resolves to the exact body version it describes. Concurrent rewrites write a *new* meta + *new* body key and never clobber the version an in-flight reader resolved. **Torn pairs are impossible by construction** (races 1 & 2), small and large objects alike. No-ETag objects fall back to the unversioned key.
- **TTL-only body lifecycle** — bodies are **never deleted synchronously** (not on overwrite, not on invalidation). Each version is immutable and ages out via TTL, so a resolved meta always finds its body and no in-flight response can be truncated by a concurrent delete. Invalidation (`DeleteWithMeta`) removes only the metadata + writes a tombstone; that makes subsequent reads miss/refetch and blocks repopulation, while the body ages out.
- **Post-PUT/DELETE re-invalidation** — `HandlePutObject`/`HandleDeleteObject` re-invalidate *after* upstream acknowledges the write. The second tombstone blocks stale repopulation a racing GET started during the write window (race 3).

## Review-driven design note
A first pass used eager eviction (delete the superseded body on overwrite) to bound storage. The high-effort review showed that reintroduced a delete-during-read truncation window (confirmed for the range/206 path) and added a `GetMeta` read to every write. Switched to the **TTL-only** lifecycle above, which eliminates all three review findings. Trade-off: superseded body versions persist until TTL — negligible for the small metadata files in #92; a background reaper is the follow-up for large-object-heavy workloads.

## Migration
None. Old unversioned bodies become unreachable and age out via TTL; those reads miss and refill — a one-time miss bump on deploy, no correctness issue.

## Tests / verification
- `cache`: meta resolves to its own body even when another version's body exists (no torn pair); versions coexist after overwrite (no sync eviction); `DeleteWithMeta` removes meta but not body; empty-ETag fallback.
- `proxy`: post-PUT re-invalidation clears a racing repopulation.
- ✅ `make test`, `make test-race` (clean), `make vet`, `make lint-ci` all pass.

Closes #92.

---

> [!NOTE]
> **High Risk**
> Changes core S3 object cache correctness and all GET/range/revalidation paths; mistakes could serve wrong bytes, truncated responses, or stale data after writes.
> 
> **Overview**
> Fixes torn meta/body pairs and read-after-write races under concurrent GETs and writes on hot keys (#92).
> 
> **ETag-versioned cache bodies** — Bodies are keyed by `body|bucket|key|<etag>`; reads use `meta.ETag` so metadata always pairs with the right bytes. Overwrites add new keys instead of clobbering in flight. **Invalidation drops metadata only** (plus tombstone); bodies age out via TTL and are not deleted on failed writes or tombstone skips, avoiding truncation of active readers. Empty-ETag objects are not cached (`IsCacheable` / `PutWithMeta`); versioned misses do not fall back to legacy unversioned body keys.
> 
> **Proxy consistency** — PUT/DELETE re-invalidate after successful upstream 2xx (`statusRecorder`); CopyObject and bulk delete use captured responses (including S3 200 + `<Error>`) so failed mutating ops do not tombstone valid racing refills. **Orphaned meta** (body evicted) is cleared and requests fall through to upstream; range-from-cache probes the body before sending 206.
> 
> Tests cover versioning guarantees, re-invalidation gating, and integration range fallthrough.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fc1c37d682e85e34f8eda4e61ac4d7e4a22dcc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -68,36 +68,12 @@ func (c *Cache) IsEnabled() bool {
 // IMPORTANT: Body is written BEFORE metadata to ensure metadata presence
 // guarantees body availability. This prevents race conditions where a reader
 // finds metadata but body hasn't been written yet.
-// currentBodyETag returns the ETag of the currently-cached metadata for the key
-// and whether such metadata exists. Used to locate a superseded body version for
-// eviction on overwrite.
-func (c *Cache) currentBodyETag(ctx context.Context, bucket, key string) (string, bool) {
-	m, ok, _ := c.GetMeta(ctx, bucket, key)
-	if !ok {
-		return "", false
-	}
-	return m.ETag, true
-}
-
-// evictSupersededBody deletes the body of a previously cached version once a new
-// version with a different ETag has been committed, bounding disk usage for keys
-// that are rewritten frequently. It runs after the new meta+body are live, so new
-// readers already resolve to the new version; an in-flight reader that resolved
-// the old meta reads its body via an atomic Get and therefore either completes
-// before this delete or cleanly misses and falls through to upstream.
-func (c *Cache) evictSupersededBody(ctx context.Context, bucket, key, prevETag string, hadPrev bool, newETag string) {
-	if !hadPrev {
-		return
-	}
-	prevBodyKey := MakeBodyKey(bucket, key, prevETag)
-	if prevBodyKey == MakeBodyKey(bucket, key, newETag) {
-		return // same ETag — body was overwritten in place, nothing to evict
-	}
-	if err := c.client.Delete(ctx, prevBodyKey); err != nil && !isNotFoundError(err) {
-		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).Msg("Failed to evict superseded body")
-	}
-}
-
+// Body lifecycle note: bodies are addressed by ETag and are never deleted
+// synchronously (not on overwrite, not on invalidation). Each version is an
+// immutable entry that ages out via TTL, so a reader that resolved a given
+// metadata version always finds its exact body — no delete-during-read can
+// truncate an in-flight response. Invalidation removes only the metadata (plus a
+// tombstone), which is enough to make subsequent reads miss and refetch.
 func (c *Cache) PutWithMeta(ctx context.Context, bucket, key string, meta *CachedObjectMeta, body []byte, ttl int) error {
 	if !c.IsEnabled() {
 		return nil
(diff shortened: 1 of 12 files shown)


ovaistariq:
@greptile

ovaistariq:
@greptile

ovaistariq:
@greptile

PR #99, opened 2026-07-16
fix: bound cache-populate concurrency by memory, not just count

ovaistariq:
## What & why

Fixes #98. Under a parquet range-read workload (Parseable), TAG pods' memory exploded from ~200 MB to **21.6 GB** in minutes, tripping `/health` liveness timeouts and pod restarts; during the restarts the k8s Service had no ready pod and returned **HTTP 503** to clients.

Root cause: every range-read cache-miss triggers a background fetch of the **entire** object, and these cache-populate operations were bounded only by **count** (`max_concurrent_writes`, default 256). Each populate buffers up to ~`channel_buffer × chunk_size` bytes (~64–80 MB via the broadcast listener channel + cache-write queue), so **256 × ~80 MB ≈ 20 GB** — the count bound is byte-unaware. Diagnosis confirmed from prod metrics: admission shed = 0, inflight = 6–9, OOM events = 0, but `tag_active_background_fetches` spiked to 828 and working-set to 21.6 GB.

## The fix

Make cache-populate admission **memory-aware**. `NewService` sizes the populate semaphore as:

```
min(max_concurrent_writes, max_populate_memory_bytes / per_populate_buffer_bytes)
```

where `per_populate_buffer_bytes ≈ (channel_buffer + channel_buffer/4) × chunk_size`.

- New config `cache.max_populate_memory_bytes` (env `TAG_CACHE_MAX_POPULATE_MEMORY`), default **1 GiB**; negative disables the memory cap (prior count-only behavior); a non-positive count still disables the limiter entirely (unchanged).
- With defaults: **~12 concurrent populates (~1 GB)** instead of 256 (~20 GB), logged at startup as `effective_cache_populate_slots`.
- Excess populates are declined exactly as before — object served from upstream, not cached. No new failure mode.
- Also corrects the `channel_buffer` default comment (1024, not 32).

## Tests

- `TestEffectiveCacheWriteLimit` (+ zero-broadcast fallback): budget reduces count (256→12), count is binding cap, tiny-budget floor of 1, negative budget/count semantics.
- Config: `max_populate_memory_bytes` default, env override (negative disables), YAML.

Build, all unit suites, race detector, and lint pass locally.

## Follow-ups (not in this PR)
- Consider not full-object-warming large objects on a single range read (the KB→GB amplification itself).
- Review `/health` liveness leniency so a memory-pressured-but-alive pod isn't killed into a crashloop.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[...]

PR #103, opened 2026-07-16
fix: match DeleteObjects results by (key, versionId), not key alone

ovaistariq:
## What & why

Cursor bot review (Medium). A `DeleteObjects` request can list the **same key multiple times with different `VersionId`s**. The post-forward re-invalidation (added in #99) matched failures by **key alone** (`failedDeleteKeys`), so if one entry for a key succeeded and another failed, the key landed in the "failed" set and was **skipped** — leaving a refill that raced the successful deletion cached as **stale data** (a read-after-write violation).

## Fix

Match delete request/response entries by **`(key, versionId)`**:
- Parse `<Error>` entries with their `VersionId`.
- Re-invalidate a key if **at least one** of its requested entries is *not* in the errored set (i.e. at least one version was actually deleted).

Errors are always listed in **both** Quiet and non-Quiet modes (only successes are omitted in Quiet mode), so "requested entry ∉ errored ⇒ succeeded" holds in both. A parse failure still over-invalidates every requested key (safe).

## Tests
- `TestHandleDeleteObjects_MixedVersionSuccessReinvalidatesKey` — same key, v1 succeeds + v2 fails ⇒ key is re-invalidated (stale refill dropped).
- Existing `TestHandleDeleteObjects_PartialFailureKeepsFailedKeyRefill` still passes (a fully-failed key keeps its refill).

Build, proxy tests (incl. race), the DeleteObjects integration test, and lint pass locally.

Refs #92.

---

> [!NOTE]
> **Medium Risk**
> Changes cache invalidation logic on a hot S3 proxy path; incorrect gating could serve stale objects or drop valid cache entries, though behavior is narrowed and covered by new tests.
> 
> **Overview**
> Fixes **post-forward cache re-invalidation** for bulk `DeleteObjects` when S3 returns partial per-object `<Error>` entries on a 200 response.
> 
> Instead of treating any error for a key as “fully failed,” the handler now compares **how many delete entries were requested per key** with **how many `<Error>` entries upstream returned for that key**. A key is re-invalidated when at least one requested entry likely succeeded (`requested > errored`); if every entry for a key errored, the racing refill is kept. **Parse failures** still re-invalidate all requested keys (safe over-invalidation).
> 
> Error counting is **by key only** (not `(key, VersionId)` tuples) so mixed success/failure for the same key under different versions re-invalidates correctly, while failed unqualified deletes with mismatched `VersionId` in the response do not over-invalidate.
> 
> Adds tests for mixed version success/failure and `VersionId` representation mismatches on failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41671b10b198cda2ea5682c4e4896f7cc673a5a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/delete_objects.go
+++ b/proxy/delete_objects.go
@@ -30,37 +30,32 @@ type deleteObjectEntry struct {
 }
 
 // deleteObjectsResult represents the S3 DeleteObjects response. Only the per-object
-// <Error> entries are captured (with their VersionId): successful deletes may be
-// omitted (Quiet mode), but errors are always listed in both modes, so a requested
-// entry succeeded iff it is NOT in the errored set.
+// <Error> entries are captured: successful deletes may be omitted (Quiet mode), but
+// errors are always listed in both modes, so a key had at least one successful
+// delete iff more entries were requested for it than upstream reported as errored.
 type deleteObjectsResult struct {
 	XMLName xml.Name `xml:"DeleteResult"`
 	Errors  []struct {
-		Key       string `xml:"Key"`
-		VersionId string `xml:"VersionId"`
+		Key string `xml:"Key"`
 	} `xml:"Error"`
 }
 
-// deleteEntryID identifies a delete request/response entry by (key, versionId).
-// A DeleteObjects request may list the same key multiple times with different
-// version IDs, so entries must be matched by both, not by key alone.
-func deleteEntryID(key, versionID string) string {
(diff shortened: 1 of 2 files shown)


ovaistariq:
@greptile

pboros (approved):

PR #104, opened 2026-07-16
fix: don't evict valid cache entries on transient body-read errors

ovaistariq:
## 1. Cache wiped on transient body errors (fixed)

The orphaned-meta invalidation ran on **every** failed cache body read, not only when the body was missing. `serveRangeFromCache` reported `served=false` for *any* probe error and **discarded the cause**, so callers deleted the metadata unconditionally. A client disconnecting mid-probe (canceled context) or a transient I/O error would tombstone a **still-valid hot entry**, forcing needless upstream refetches — under a burst of disconnects this can wipe hot keys.

**Fix:** `serveRangeFromCache` now returns the underlying probe error instead of `nil`, and a new `bodyGone(err)` predicate gates every invalidation site (range path, full-object path, `forwardAfterCacheMiss`). Only `cache.ErrNotFound` and the new `errCacheBodyEmpty` sentinel — the genuinely orphaned-metadata cases the invalidation exists to heal — evict. Canceled contexts and I/O errors leave the entry alone.

Adds `TestBodyGone` (missing/empty ⇒ evict; canceled/deadline/io-error ⇒ don't).

## 2. Copy success ignores incomplete capture (documented, behavior intentionally unchanged)

I looked at this closely and **did not change the behavior**, because the suggested direction would trade a perf issue for a correctness bug. The two failure modes are **not symmetric**:

- Treating a **failed** copy as successful ⇒ re-invalidates a destination that didn't change ⇒ **an extra cache miss** (perf only).
- Treating a **successful** copy as failed ⇒ skips invalidation ⇒ a racing refill of the **old** destination stays cached ⇒ **stale data** (correctness).

So gating on `capture.Complete` (i.e. treating an incomplete capture as "not succeeded") would risk serving stale data whenever a *successful* copy's capture was truncated. An unconfirmable outcome must be assumed successful.

Also worth noting the flagged window is narrower than it looks: `isS3ErrorBody` only reads the **first** start element, so an incomplete capture still detects a failure whenever the `<Error>` root was captured. Only a body truncated *before* the root reads as success — which is the safe side.

Added a comment recording this reasoning so it isn't re-litigated.

Build, all unit suites, race, integration, and lint pass locally.

Refs #92.

---

> [!NOTE]
> **Medium Risk**
> Changes cache invalidation semantics on GET/revalidation paths; incorrect classification could either prolong orphaned-meta loops or evict valid entries, but the change is narrow, tested, and defaults to healing on ambiguous errors.
> 
> **Overview**
> Stops **unconditional metadata deletion** when a cache body read fails. Previously any probe failure (including client disconnect / canceled context) could evict a still-valid hot entry and force repeated upstream fetches.
> 
> Adds **`bodyGone(err)`**, a denylist of transient failures (`context.Canceled`, `DeadlineExceeded`, gRPC `Canceled` / `DeadlineExceeded` / `Unavailable`). Everything else is treated as a genuinely missing or inconsistent body and triggers heal-by-delete. **`serveRangeFromCache`** now returns the underlying probe error instead of `nil`, so callers can distinguish transient vs orphaned-meta cases.
> 
> **Invalidation is gated** on `bodyGone` for range hits, full-object cache serves, and **`forwardAfterCacheMiss`** (revalidation / stale fallbacks). Orphaned-meta healing still runs for not-found, EOF, invalid range, empty body, etc.
> 
> Adds **`TestBodyGone`** and **`TestServeRangeFromCache_EmptyBodyReportsGone`**. Documents why **`copyObjectSucceeded`** does not require `capture.Complete` (stale-data risk vs extra miss); behavior unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e15bfc6193dbc74dc3451802fc62b436184e222d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/body_gone_test.go
+++ b/proxy/body_gone_test.go
@@ -4,14 +4,24 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"testing"
+	"time"
 
+	cacheclient "github.com/tigrisdata/ocache/client"
 	"github.com/tigrisdata/tag/cache"
+	"github.com/tigrisdata/tag/config"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
-// TestBodyGone verifies only a genuinely missing/inconsistent body invalidates the
-// orphaned metadata. Transient failures — most importantly a canceled context from
-// a client that disconnected mid-read — must NOT evict a still-valid hot entry.
+// TestBodyGone pins the denylist contract: only transient failures (where the
+// cached body is probably fine and we merely couldn't finish reading it) skip the
+// heal. Every other error invalidates the orphaned metadata — the cache reports an
+// unusable body in several shapes, and missing one leaves meta outliving its body,
+// which forwards every request upstream until the meta TTL expires.
 func TestBodyGone(t *testing.T) {
 	tests := []struct {
 		name string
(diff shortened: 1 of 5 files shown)


ovaistariq:
@greptile

Issue #97, opened 2026-07-15
fix: tighten read-after-write guarantee — tombstone re-invalidation is not airtight


ovaistariq:
## Context

PR #95 added post-forward re-invalidation to PUT/DELETE/Copy/bulk-delete to close the read-after-write race where a GET that raced an in-flight write re-caches the pre-write object. This substantially narrows the window but does **not** fully close it. Two inherent weaknesses remain in the tombstone mechanism.

## [1] `writeStartTime` is stamped after the upstream read snapshot

The tombstone-aware cache write guards on `tombTs >= writeStartTime` (`cache/cache.go` `PutWithMetaStreamTombstoneAware`). But `writeStartTime` is captured in `setupCacheListener` (`proxy/caching.go:116`), which runs **after** `DoRequestWithCreds` returns response headers (`proxy/get_object.go:300` → `:317`). So the timestamp reflects when the proxy *began the cache write*, not the moment upstream took the read snapshot.

If upstream serves the pre-write version but the write lands and its tombstone is written before the GET stamps `writeStartTime`, the guard `tombTs >= writeStartTime` is false and the stale repopulation proceeds. Whether this window is reachable depends on upstream (Tigris) read/write ordering semantics for concurrent connections; it is narrow but not provably empty.

## [2] Tombstone TTL (60s) is shorter than large-object writes

`tombstoneTTL = 60` (`cache/cache.go`). A legitimate large-object cache write can run far longer than 60s (`cacheWriteTimeoutForSize` scales with size, `proxy/caching.go:74`). If a write's tombstone is written at the start and the racing populate finishes >60s later, `GetTombstoneTimestamp` returns 0 (expired), the guard fails, and stale pre-write content is cached.

## Possible directions

- Stamp `writeStartTime` from an **upstream-provided read-snapshot timestamp** (e.g. `Date` / `x-amz-*` header on the GET response) rather than proxy wall-clock, so the compare is against when upstream actually read.
- Capture `writeStartTime` **before** `DoRequestWithCreds` rather than after.
- Scale tombstone TTL to at least the max cache-write timeout, or refresh the tombstone while a write is in flight.
- Consider a generation/version counter per key instead of timestamp comparison.

## Notes

Both are **pre-existing** limitations of the tombstone design, not regressions introduced by #95. Filed as a design follow-up to make read-after-write provably tight. Related: #95, #92, #96.

ovaistariq:
[...]

PR #106, opened 2026-07-17
fix: tighten read-after-write by stamping writeStartTime before the upstream read

ovaistariq:
Implements **A + B** from #97. Both were pre-existing holes in the tombstone guard, not regressions.

## [A] `writeStartTime` was stamped after the upstream read

The guard is `tombTs >= writeStartTime`, but upstream takes its read snapshot *after* we send. Stamping when the headers arrive left the **entire round-trip unguarded**:

```
T_send   GET sent
T_snap   upstream snapshot        -> returns PRE-PUT data
T_put    a PUT lands upstream
T_tomb   its tombstone written    <- inside our upstream latency
T_hdr    headers arrive           -> writeStartTime stamped HERE
```
`T_tomb < T_hdr` ⇒ guard is **false** ⇒ we cache the pre-PUT body we just read. Stale, for the full cache TTL. The window is the GET's upstream latency — not microseconds.

**Fix:** stamp before issuing the request; thread it into `setupCacheListener` (inline GET + background warm). Now `T_ws < T_snap`, so:
- read pre-PUT data ⇒ `T_ws < T_snap < T_put < T_tomb` ⇒ guard fires ⇒ **stale write always skipped** ✅
- read fresh data ⇒ may skip a populate ⇒ **a cache miss, never staleness**

A deliberate one-directional trade: an occasional self-healing miss instead of a correctness hole.

**Intentionally untouched:** `handleRevalidation200` stamps *after* its own `Delete` so its own tombstone doesn't block its write; stamping earlier would make every revalidation-200 skip its cache update. That sub-case needs its own design and stays open on #97.

## [B] Tombstone TTL was ~5x too short

Fixed `60s`, while a populate runs up to `cacheWriteTimeoutForSize(size_threshold)` (**205s** for 1 GiB) plus a fetch up to `backgroundFetchTimeout` (**300s**). The tombstone expired mid-write ⇒ guard read zero ⇒ stale write proceeded. Fix [A] *widens* that window (it now includes the round-trip), so this had to move with it.

TTL is now **derived from `size_threshold`** (floor 10 min), so raising `size_threshold` can't silently reintroduce the race. Lengthening is nearly free: a tombstone only blocks writes whose `writeStartTime` precedes it, so later writes are never blocked — cost is an 8-byte entry living longer.

## Tests

- **`TestStreamFromUpstream_TombstoneDuringFetchBlocksPopulate`** — drives a real GET whose upstream fetch invalidates mid-flight; asserts the populate is blocked *and* the client still gets its bytes, with a control proving the harness caches when there's no race. **Verified it fails when the old stamp order is reintroduced** (`stale populate was not blocked`) — it genuinely catches the bug.
- **`TestTombstoneTTLCoversPopulateWindow`** — cross-package drift guard: TTL must outlive `cacheWriteTimeoutForSize + backgroundFetchTimeout` across thresholds, so moving either side can't silently reopen the race.
- **`TestTombstoneTTLSeconds`** — floor, scaling, monotonicity.

Build, all unit suites, race, integration, and lint pass locally.

Refs #97.

---

> [!NOTE]
> **High Risk**
> Changes core cache invalidation and populate correctness; mistakes could serve stale data for the full object TTL, though behavior is tightened and covered by new race/TTL tests.
> 
> **Overview**
> Closes two tombstone-guard holes that could **cache stale object bodies** after a racing PUT/DELETE (#97).
> 
> **`writeStartTime` before upstream:** Inline GET populates and background warms now record the cache-write start **before** the upstream request and pass it into `setupCacheListener`. Invalidations during the fetch produce tombstones newer than the stamp, so `PutWithMetaStreamTombstoneAware` skips the metadata write instead of caching pre-invalidation bytes. Clients still get the streamed response; only populate may be skipped (extra miss, not staleness).
> 
> **Tombstone TTL from `size_threshold`:** Invalidation markers no longer use a fixed 60s TTL. `TombstoneTTLSeconds` derives lifetime from `cache.size_threshold` (10-minute floor, scaled by assumed 5 MB/s write throughput × safety factor) so tombstones outlive long populates and raising the threshold does not shorten the guard mid-write.
> 
> **Tests:** Integration-style GET test with mid-fetch tombstone, cross-package TTL vs populate-window check, and unit tests for TTL scaling/monotonicity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b48cb11c7497b6e343e1a7b80755468d5a7b626f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -22,10 +22,11 @@ var ErrCacheDisabled = errors.New("cache is disabled")
 
 // Cache wraps ocache client for TAG.
 type Cache struct {
-	client     cacheclient.CacheClient
-	defaultTTL int64 // seconds
-	enabled    bool
-	closed     bool
+	client       cacheclient.CacheClient
+	defaultTTL   int64 // seconds
+	tombstoneTTL int64 // seconds; must outlive the longest racing cache-populate
+	enabled      bool
+	closed       bool
 }
 
 // NewCacheWithClient creates a cache with an injected client.
@@ -33,16 +34,19 @@ type Cache struct {
 func NewCacheWithClient(client cacheclient.CacheClient, cfg *config.CacheConfig) *Cache {
 	ttl := int64(config.DefaultCacheTTL.Seconds())
 	enabled := true // Default to enabled
+	var sizeThreshold int64
 	if cfg != nil {
 		if cfg.TTL > 0 {
 			ttl = int64(cfg.TTL.Seconds())
 		}
 		enabled = cfg.IsEnabled()
+		sizeThreshold = cfg.SizeThreshold
 	}
 	return &Cache{
-		client:     client,
-		defaultTTL: ttl,
-		enabled:    enabled,
+		client:       client,
+		defaultTTL:   ttl,
+		tombstoneTTL: TombstoneTTLSeconds(sizeThreshold),
+		enabled:      enabled,
 	}
 }
 
@@ -50,7 +54,8 @@ func NewCacheWithClient(client cacheclient.CacheClient, cfg *config.CacheConfig)
 // All operations return successfully with "not found" or nil results.
 func NewDisabledCache() *Cache {
 	return &Cache{
-		enabled: false,
+		enabled:      false,
+		tombstoneTTL: MinTombstoneTTLSeconds,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


ovaistariq:
Fixed in the latest push — the finding was correct.

**Verified the collapse.** The TTL was `max(600, 2*write)` (multiplicative) while the window it must outlive is `max(60, write) + 300` (additive). Those curves cross. Margin by threshold:

| size_threshold | TTL | window | margin |
|---|---|---|---|
| 1.0 GiB | 600s | 505s | 95s |
| **1.46 GiB** | **600s** | **599s** | **1s** |
| **1.5 GiB** (`write == 300`) | **600s** | **600s** | **0** |
| 10 GiB | 4096s | 2348s | 1748s |

At `write == 300` the floor and the derived value both land on 600s and the tombstone no longer strictly outlives the populate — exactly as described.

**Fix:** derive the TTL from the window's own shape (`write + fetch + margin`) rather than scaling write time by a factor. Margin is now a constant 300s at every threshold.

**On the test that let this through:** my three thresholds (64 MiB, 1 GiB, 10 GiB) had margins of 96s/95s/1748s — all straddling the collapse band without landing in it. Spot-checking two independently-growing curves cannot find where they cross, so `TestTombstoneTTLCoversPopulateWindow` now sweeps 0 → 24 GiB in 16 MiB steps and asserts `ttl > window` throughout. Confirmed it fails against the old formula:

```
size_threshold=1577058304 (1.47 GiB): tombstone TTL 10m0s does not outlive the populate window 10m0s
--- FAIL: TestTombstoneTTLCoversPopulateWindow
```

and passes against the new one, reporting a smallest margin of 5m0s. `TestTombstoneTTLSeconds` now pins additivity directly (doubling the threshold must add the extra write time, not double the TTL) so a multiplicative curve cannot come back.

ovaistariq:
@greptile

PR #111, opened 2026-07-17
fix: invalidate read cache on CompleteMultipartUpload

### LINKED ISSUE #110: fix: CompleteMultipartUpload does not invalidate the read cache, serving stale bytes after a multipart overwrite
## Summary

`HandleCompleteMultipartUpload` (`proxy/service.go:434`) is the only object-mutating path that does **not** invalidate the object's read cache. When a multipart upload completes over a key whose previous version is cached, TAG can serve the **old** object bytes on a subsequent GET.

## Evidence

Surfaced by the ceph s3-test `test_multipart_upload_overwrite_existing_object`, which:

1. `put_object(key, 5 MiB)`
2. multipart-uploads 2×5 MiB over the same key and completes it (new object = 10 MiB)
3. `get_object(key)` and asserts the body equals the 10 MiB payload

The GET returned the **5 MiB** first version. pytest's diff makes it unambiguous — the two strings share `5242869` (~5 MiB) identical leading characters and then the old object simply ends:

```
E  AssertionError: assert '123451234512...' == '123451234512...'
E    Skipping 5242869 identical leading characters in diff
FAILED test_s3.py::test_multipart_upload_overwrite_existing_object
```

Intermittent (surfaces when the prior version is cached and, in cluster mode, when routing widens the window). It is **not** related to the read-after-write PRs (#92/#95/#97/#106) — those don't touch this handler — but it is the same *class* of bug on a path they didn't cover.

## Root cause

Every other mutation path invalidates the read cache before forwarding **and** re-invalidates after a 2xx (writing a tombstone newer than any racing populate's start time, restoring read-after-write semantics):

- `HandlePutObject` — `service.go:235` + `:253`
- `HandleDeleteObject` — `service.go:~273` + `:~284`
- `HandleCopyObject` (destination) — `service.go:~382` + `:~394`

`HandleCompleteMultipartUpload` does neither. It checks the completion-idempotency cache, forwards with capture, and caches the completion response — but never calls `s.cache.Delete(bucket, key)`. The stale `meta|bucket|key` + versioned body from the pre-multipart object survive, and a later GET resolves them.

## Proposed fix

Mirror the PUT/DELETE/COPY pattern in `HandleCompleteMultipartUpload`:

1. Invalidate `s.cache.Delete(ctx, bucket, key)` **before** forwarding to upstream.
2. Re-invalidate **after** upstream confirms a 2xx completion, so a GET that raced the in-flight completion and began re-caching the pre-overwrite object is blocked by a newer tombstone.

Notes / care points:
- Keep the existing completion-idempotency caching (`GetCompletion`/`PutCompletion`) untouched — that's keyed on `uploadId` and is a different concern from the object's read cache.
- The object key here is the final object being written by the multipart completion, so invalidating `(bucket, key)` is correct; don't invalidate on `CreateMultipartUpload`/`UploadPart` (the object doesn't change until completion).

## Verification

- `test_multipart_upload_overwrite_existing_object` passes deterministically (single-node **and** cluster).
- Add a proxy unit test: cache a version of a key, run CompleteMultipartUpload over it, assert the meta is invalidated (tombstone written) and a subsequent GET refetches.

ovaistariq:
Closes #110.

## Problem

`HandleCompleteMultipartUpload` was the only object-mutating path that never invalidated the object's **read** cache — it forwarded to upstream and cached the completion response for idempotent replays, but wrote no tombstone and deleted no meta. Completing a multipart upload overwrites the object, so a previously cached version survived and a later GET could serve the **old** bytes.

Surfaced (intermittently) by the ceph s3-test `test_multipart_upload_overwrite_existing_object`: after overwriting a 5 MiB object with a 10 MiB multipart upload, the GET returned the old 5 MiB object. Same class of read-after-write bug as #92 / #95 / #97, on the path they didn't cover.

## Fix

Mirror the `PutObject` / `DeleteObject` / `CopyObject` pattern in `HandleCompleteMultipartUpload`:

- **Invalidate before forwarding** so a forward that succeeds but whose post-invalidation fails can't leave stale data served.
- **Re-invalidate after a confirmed-successful completion** so a GET that raced the in-flight completion and re-cached the pre-overwrite object is dropped.
- The **idempotent-replay path** (completion served from the completion cache) changes nothing about the object and needs no invalidation — the first completion already invalidated it.

CompleteMultipartUpload shares CopyObject's "200 OK with `<Error>` body" failure mode, so the re-invalidation is gated on the same success check. Renamed `copyObjectSucceeded` → `s3WriteSucceeded` now that both paths use it. As a corollary, the completion is only cached for idempotent replay on a genuine success — previously a 200-with-error-body would have been stored and replayed as success.

## Tests

`proxy/multipart_complete_test.go`:
- success invalidates the overwritten object (incl. a racing mid-completion refill)
- 200-with-`<Error>` body keeps a racing refill and is not cached as a successful completion
- pre-forward invalidation fires even when the forward errors
- a successful completion is still cached for idempotent replay (second identical call served without re-forwarding)

`make lint`, `make test-proxy`, `make test-cache` all green.

---

> [!NOTE]
> **Medium Risk**
> Changes core cache consistency on object-mutation paths (multipart complete, bulk delete, invalidation metrics); incorrect gating on `s3WriteSucceeded` could still serve stale data or drop valid cache entries.
> 
> **Overview**
> Fixes read-after-write staleness when **CompleteMultipartUpload** overwrites an object: the handler now **invalidates read cache before forwarding** and **re-invalidates after a confirmed success**, matching Put/Delete/Copy. Idempotent completion replays from the completion cache still skip invalidation.
> 
> **`s3WriteSucceeded`** (renamed from `copyObjectSucceeded`) gates post-write invalidation and completion idempotency caching for both Copy and multipart complete, so **200 OK with an S3 `<Error>` body** is not treated as a successful write and is not stored for replay.
> 
> **`invalidateObject`** centralizes pre-forward invalidation for Put, Delete, bulk delete, and multipart complete; **delete metrics record error** when the cache backend fails instead of always reporting success.
> 
> **`DeleteWithMeta`** still runs tombstone + meta delete best-effort but **returns joined errors** when either backend step fails (not-found meta delete remains success). New cache and proxy tests cover failure propagation and multipart invalidation edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dddcd52e4b9965a32294f1b81049e797be0da64a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"strings"
 	"time"
@@ -289,15 +290,25 @@ func (c *Cache) GetBodyStream(ctx context.Context, bucket, key, etag string, w i
 
 // DeleteWithMeta removes both metadata and body from cache.
 // Writes a tombstone first to prevent in-flight cache writes from completing.
+//
+// Both steps are attempted even if the first fails (best-effort invalidation), but a
+// genuine backend failure of either is returned so callers don't report a successful
+// invalidation while stale metadata is still readable. A not-found metadata delete is
+// success — the entry is already gone.
 func (c *Cache) DeleteWithMeta(ctx context.Context, bucket, key string) error {
 	if !c.IsEnabled() {
 		return nil
 	}
 
-	// Write tombstone FIRST - prevents in-flight writes from completing
+	var errs []error
+
+	// Write tombstone FIRST - prevents in-flight writes from completing. A failure
+	// here leaves the invalidation incomplete (an in-flight populate could resurrect
+	// the entry), so it is a real failure — but still continue to the meta delete.
 	if err := c.WriteTombstone(ctx, bucket, key); err != nil {
 		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).
 			Msg("Failed to write tombstone (continuing with delete)")
+		errs = append(errs, fmt.Errorf("write tombstone: %w", err))
 	}
 
 	metaKey := MakeMetaKey(bucket, key)
@@ -310,6 +321,11 @@ func (c *Cache) DeleteWithMeta(ctx context.Context, bucket, key string) error {
 	// streaming that exact version.
 	if err := c.client.Delete(ctx, metaKey); err != nil && !isNotFoundError(err) {
 		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).Msg("Cache meta delete error")
+		errs = append(errs, fmt.Errorf("delete meta: %w", err))
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
 	}
 
 	log.Debug().Str("bucket", bucket).Str("key", key).Msg("Invalidated cache metadata (body ages out via TTL)")
--- a/cache/delete_error_test.go
+++ b/cache/delete_error_test.go
@@ -1,0 +1,87 @@
+package cache
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	cacheclient "github.com/tigrisdata/ocache/client"
+	"github.com/tigrisdata/tag/config"
+)
+
+// flakyClient wraps a real client and can inject failures into the two backend
+// operations DeleteWithMeta performs: Put (tombstone) and Delete (metadata).
+type flakyClient struct {
+	cacheclient.CacheClient
+	putErr    error // returned by Put (tombstone write) when non-nil
+	deleteErr error // returned by Delete (metadata delete) when non-nil
+}
+
+func (f *flakyClient) Put(ctx context.Context, key string, data []byte, ttlSeconds int64) error {
+	if f.putErr != nil {
+		return f.putErr
+	}
+	return f.CacheClient.Put(ctx, key, data, ttlSeconds)
+}
+
+func (f *flakyClient) Delete(ctx context.Context, key string) error {
+	if f.deleteErr != nil {
+		return f.deleteErr
+	}
+	return f.CacheClient.Delete(ctx, key)
+}
+
+func newCacheWithClientForTest(t *testing.T, client cacheclient.CacheClient) *Cache {
+	t.Helper()
+	cfg := config.NewDefault()
+	return NewCacheWithClient(client, &cfg.Cache)
+}
+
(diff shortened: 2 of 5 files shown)


ovaistariq:
Fixed in the latest push (`fix: record cache-invalidation failures instead of reporting success`).

The observation is correct — `Cache.Delete`'s error was discarded and `("delete","success")` recorded unconditionally, so a backend failure would leave stale metadata while monitoring showed a green delete. Worth noting it was a pre-existing pattern shared by every metric-recording invalidation site (`PutObject`, `DeleteObject`, bulk `DeleteObjects`), not new to this PR — my CompleteMultipartUpload block was copied verbatim from them. So rather than fix only the new site and leave the identical siblings wrong, I routed all four through a shared `Service.invalidateObject` that records `("delete","error")` and debug-logs on failure, and `("delete","success")` only on an actual success.

The `error` label is additive, so existing success dashboards are unaffected except that a genuine failure now stops inflating the success count. The post-forward re-invalidations are unchanged — they deliberately don't record this metric to avoid double-counting the same logical invalidation.

ovaistariq:
@greptile

ovaistariq:
Good catch — the previous commit was incomplete. `Cache.Delete` → `DeleteWithMeta` logged and swallowed both the tombstone-write and metadata-delete errors and always returned `nil`, so my helper's `("delete","error")` branch was unreachable and it still always recorded success.

Fixed at the source: `DeleteWithMeta` now attempts both steps (best-effort invalidation — a tombstone failure still proceeds to the meta delete) but accumulates and returns a joined error when either genuinely fails. A not-found metadata delete stays successful, since the entry already being gone is the goal state. `invalidateObject` is the only caller that consumes the return value; all others already ignore it, so nothing else changes behavior.

Added cache-package tests (`cache/delete_error_test.go`) via a client wrapper that injects backend failures:
- tombstone `Put` fails → `DeleteWithMeta` returns an error
- metadata `Delete` fails → returns an error
- healthy invalidation → nil, and the meta is actually gone
- `"key not found"` on the meta delete → nil (not-found is success)

`make lint`, `test-cache`, `test-proxy` all green.

ovaistariq:
@greptile

Issue #109, opened 2026-07-17
Time based eviction


pboros:
We are using TAG for parquet files that are written once (the most recently written data is read the most). Instead of LRU, can we have time based eviction? Maybe time and size based, and also have a way to tell the oldest cached entry in the system. With LRU, a read on older parquet files, which are very rare can evict otherwise valuable cache for the generic workload.

ovaistariq:
Filed on the ocache side as tigrisdata/ocache#185 (time-based / FIFO eviction policy).

**Where it stands today:** ocache eviction is LRU by last-access and only runs under `-max-disk-usage`. Because a read refreshes an entry's recency, a rare read of an old parquet file protects it and evicts more valuable generic-workload data — exactly the problem you described.

**Proposed fix (in ocache#185):** add `-eviction-policy=lru|fifo`. In `fifo` mode, evict oldest-*written*-first by simply not refreshing the timestamp on reads — the write path already stamps insertion time and the eviction loop already walks oldest-first, so it's a small, self-contained change (no proto/format changes). Pairing FIFO with `-ttl` gives the "time and size based" variant you mentioned.

**One scoping note:** the "tell me the oldest cached entry" ask is tracked separately as ocache#183 (an `ocache_oldest_entry_age_seconds` metric) and is currently deferred — it isn't needed for the eviction change, so FIFO eviction can land independently. If surfacing the oldest-entry age is still valuable to you, let us know and we can prioritize that half too.

ovaistariq:
Partially addressed by #112 (merged as baca028):

**Done — time-based eviction.** ocache v1.8.0 added a FIFO (oldest-written-first) eviction policy, now exposed via `cache.eviction_policy` / `TAG_CACHE_EVICTION_POLICY` (`lru` default, or `fifo`). FIFO is the behavior this issue asked for: for write-once parquet, a rare read of an old object no longer refreshes its recency and keeps it resident over newer, hotter data. Note it only takes effect under a disk cap (`max_disk_usage_bytes > 0`) — with no cap nothing is evicted regardless of policy.

**Still open — "a way to tell the oldest cached entry."** ocache does not yet expose an oldest-entry age metric/getter (tracked in tigrisdata/ocache#183 item 2). The v1.8.0 ordered eviction indexes make it cheap to derive, but it is not surfaced yet. Keeping this issue open scoped to that remaining piece; will wire it into TAG once ocache exposes it.

ovaistariq:
[...]

PR #112, opened 2026-07-19
feat: bump ocache to v1.8.0, add FIFO eviction option and cache-size metric

### LINKED ISSUE #108: Cache size metric
It would be great to have the cache size as a metric for cluster sizing purposes, also maybe the oldest entry in cache.

ovaistariq:
Bumps ocache v1.7.0 → v1.8.0 and surfaces two capabilities it adds. Resolves #108 and the eviction half of #109.

## Eviction policy (#109)

ocache 1.8.0 added a FIFO eviction policy (evict oldest-written first). This exposes it through TAG:

- New `cache.eviction_policy` config / `TAG_CACHE_EVICTION_POLICY` env var: `lru` (default) or `fifo`, wired to `embeddedCfg.Storage.EvictionPolicy`.
- FIFO suits write-once workloads (dated parquet) where a rare read of an old object shouldn't keep it resident at the expense of newer, hotter data.
- **Eviction only runs under a disk cap** (`max_disk_usage_bytes > 0`) — documented in `configuration.md`; with no cap nothing is evicted regardless of policy.

## Cache size metric (#108)

- New `tag_cache_size_bytes` gauge — this node's logical cache size (sum of stored object lengths), sampled every 30s from ocache's newly-exported `Storage().TotalSize()` (an atomic ocache maintains live; cheap, no rescan). Per-node; `sum(tag_cache_size_bytes)` for a cluster total.
- v1.8.0 also refreshes the equivalent `ocache_disk_usage_bytes{type="total"}` live (it was startup-only before), so that's now usable directly too — noted in `metrics.md`.

## Tests

- `config`: `eviction_policy` default (lru), YAML value, and env override (case-insensitive).
- `metrics`: `SampleCacheSize` publishes the initial value immediately and stops on context cancel.
- `go.mod` gains `client_model` as a direct dep (already transitively present) so the test reads the gauge without pulling in `testutil` (which would add `godebug`).

`make build`, `make lint`, `make test` all green.

## Scope / follow-ups

- The **oldest-entry** ask (in both #108 and #109) is still not exposed by ocache — tracked in tigrisdata/ocache#183. #109 stays open for that piece.

---

> [!NOTE]
> **Low Risk**
> Changes are additive config, metrics, and dependency bump with validation and careful shutdown ordering; no auth or request-path behavior changes.
> 
> **Overview**
> Upgrades **ocache** from v1.7.0 to **v1.8.0** and wires two new operator-facing capabilities.
> 
> **FIFO eviction** — Adds `cache.eviction_policy` / `TAG_CACHE_EVICTION_POLICY` (`lru` default, `fifo` for oldest-written-first under a disk cap). Values are normalized, validated at startup (invalid policies fail fast instead of silently falling back in ocache), and passed through to embedded storage config. Docs note eviction only applies when `max_disk_usage_bytes > 0`.
> 
> **Cache size observability** — Introduces **`tag_cache_size_bytes`** (per-node logical cache size), updated immediately and every 30s via `SampleCacheSize` from `Storage().TotalSize()`, with shutdown ordering so sampling stops before cache close.
> 
> Config tests cover eviction defaults, YAML/env overrides, invalid values, and blank env; metrics tests cover the sampler lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49ff9269f5de4085a9121b05e27618d2dd464f95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ovaistariq:
Both bot findings were valid and are fixed in the latest push.

**1. Invalid eviction policy silently applied** (Greptile P1 + Cursor). Correct — an unrecognized value (`fif0`, etc.) was forwarded to ocache, which falls back to LRU with only a warning, so the operator’s intended policy silently never applied. Cursor also correctly caught that a whitespace-only `TAG_CACHE_EVICTION_POLICY` trimmed to empty and could wipe a valid YAML/default value. Fixes:
- The env override now ignores blank/whitespace input (`strings.TrimSpace(...) == ""` → skip), so it can’t wipe YAML/defaults — matching how the sibling cache-tuning overrides ignore invalid input.
- `validate()` now rejects any `cache.eviction_policy` other than `lru`/`fifo` with a clear startup error, so a typo fails fast instead of degrading silently. Tests added for invalid-from-YAML, invalid-from-env, and blank-env-keeps-YAML.

**2. Sampler outlives cache storage** (Greptile P1). Correct — the sampler ran on the root ctx whose `cancel` is deferred at line 181, before `defer embeddedCache.Close()` (256); defers are LIFO, so `Close()` ran first and a tick could call `Storage().TotalSize()` on closing storage. Fixed by giving the sampler its own context, and cancelling + joining it (`WaitGroup.Wait()`) in a defer registered right after the sampler starts — which is after the `Close()` defer, so it runs before `Close()`. The sampler is now fully drained before the cache closes.

`make build`, `make lint`, `make test` all green.

ovaistariq:
@greptile

Issue #98, opened 2026-07-16
fix: cache-populate concurrency is byte-unaware (256 × ~64MB buffers) — memory blowup + liveness restarts under parquet range-read fan-out


ovaistariq:
## Summary

Under a parquet range-read workload (Parseable), TAG pods' memory exploded from ~200 MB to **21.6 GB** in minutes, causing `/health` liveness-probe timeouts and pod restarts. During the restarts the k8s Service had no ready pod for some requests and returned **HTTP 503** to the client. This was misattributed to TAG returning S3 `SlowDown`, but TAG generated none.

## Evidence (from metrics/logs)

- `tag_admission_shed_total` = **0** for 7 days; `tag_inflight_requests` peaked at **6–9** (limit 1024) → **not** ingress admission.
- Zero `"code":"SlowDown"` logs and zero `"Upstream response"` `status:503` logs → 503 was **not** generated by TAG nor forwarded from upstream.
- `container_oom_events_total` = **0**; peak working-set 21.6 GB < 32 GB limit → **not** an OOMKill.
- `tag_active_background_fetches` spiked **0 → 254 → 828** at 09:36–09:50, while inbound requests stayed at 6–9 and broadcasts at 0–5.
- Memory: ~200 MB → **21.6 GB/pod**, then a drop to 0.11 GB (restart), repeated. Liveness probe is `/health` timeout 5s × threshold 3.

## Root cause

Every range-read cache-miss (`handleRangeWithBackgroundCache`) triggers a **background fetch of the entire object** (`triggerBackgroundCacheFetch` → `fetchFullObjectToCache`) to warm the cache. A few-KB parquet column-chunk read amplifies into a full multi-MB/GB object download.

These background fetches are bounded by **count** (`max_concurrent_writes`, default **256**, non-blocking `acquireCacheSlot`), but **not by bytes**. Each concurrent cache-populate buffers up to ~**64–80 MB** (broadcast listener channel `ChannelBuffer=1024 × ChunkSize=64KB` = 64 MB, plus the ~16 MB `chunkQueue` in `setupCacheListener`).

**256 × ~80 MB ≈ 20 GB** — matching the observed 21.6 GB spike. The count-based bound (#72) does not cap memory because per-op buffering is large and byte-unaware.

Note: `broadcast.channel_buffer` default is **1024** (config comment at `config.go:198` incorrectly says 32; `DefaultBroadcastChannelBuffer = 1024`).

## Fix

[...]

PR #113, opened 2026-07-21
feat: add cache-warm-on-write (opt-in) for read-after-write hotness

### LINKED ISSUE #107: Caching on writes
Right now, TAG is caching on read, can it cache on write?

ovaistariq:
Implements Option B from the #107 design discussion. Closes #107.

## What

New `cache.warm_on_write` / `TAG_CACHE_WARM_ON_WRITE` (default **false**): after a successful `PutObject`, `CompleteMultipartUpload`, or `CopyObject`, TAG triggers a background full-object fetch that repopulates the cache — so a read soon after a write is a hit.

This is **cache-warm-on-write** (write-around plus async warming), **not strict write-through**: the write still invalidates, and the warm is a separate best-effort background GET. (For the taxonomy discussion — Alluxio, write-through vs. warm-on-write, and why a write-through tee would require in-memory buffering — see the thread on #107.)

## Why it's small

It reuses the existing read-path warm machinery (`triggerBackgroundCacheFetch` → `fetchFullObjectToCache`), which already provides everything the hard part needs:

- **dedup** (one in-flight warm per key), **detached timeout context** (zero client latency)
- **byte-budget + slot shedding** under pressure — the exact bound that fixed the #98/#100 warm-storm OOM
- **tombstone-safe** `writeStartTime` stamped before the GET (an invalidation racing the warm is provably newer and blocks it)
- **cheap oversized-skip** (one header round-trip, no body transfer)

So the change is: a config flag, a small `warmOnWrite(r, bucket, key)` helper, and three call sites (right after the existing post-write invalidation, on confirmed success only).

## Consistency

For `PutObject`: pre-forward tombstone → forward → post-forward tombstone (T1) → warm stamps `writeStartTime` W > T1 → warm write is newest and wins; a DELETE racing the warm lands a tombstone > W and blocks it. Read-after-write hot **and** delete-after-write safe.

## Credentials

The warm uses `ValidateAndGetCredentials` — the same source the read-path warm uses: TAG's own credentials in transparent mode, the authorized client's in signing mode. Anonymous/unvalidated writes yield no usable creds and are not warmed.

## Notes

- **Multipart** is the case a write-through tee could never cover (TAG never sees the assembled body), so warm-on-write is the only way to make those objects hot — covered here.
- Defaults **off**: it costs one extra upstream GET per write. Enabling is an explicit operator trade (bandwidth for read-after-write hotness), suited to write-then-read pipelines (e.g. #109's parquet workload).
- New metric `tag_warm_on_write_triggered_total`; `configuration.md` and `metrics.md` updated.

## Tests

- proxy: PUT enabled → warms + caches; disabled → no warm; failed PUT → no warm; no credentials → no warm; CompleteMultipartUpload enabled → warms.
- config: default false, YAML enable, env override both directions.

`make build`, `make lint`, `make test` all green.

---

> [!NOTE]
> **Medium Risk**
> Touches post-write cache behavior and anonymous vs signed warm paths; mitigated by default-off config, tombstone checks, and reuse of existing background-fetch limits.
> 
> **Overview**
> Introduces **opt-in cache-warm-on-write** via `cache.warm_on_write` / `TAG_CACHE_WARM_ON_WRITE` (default **false**). After a confirmed-successful **PutObject**, **CopyObject**, or **CompleteMultipartUpload**, TAG still invalidates as today, then may fire a **best-effort background GET** through the existing populate path (dedup, populate budget, tombstone-aware `writeStartTime`) so a read soon after a write is more likely to hit cache.
> 
> **Authenticated writes** warm with the write’s credentials; **anonymous writes** use a new **unsigned** upstream GET (`DoAnonymousFullObjectRequest` + vhost-style `anonymousObjectURL`) so public-read is learned from a real anonymous read—**403 leaves nothing cached**; **200** can cache as public-read. Adds `tag_warm_on_write_triggered_total`, docs, config tests, proxy/integration coverage, and `RequestSigner.Endpoint()` for URL building.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec696fad703678dcd5bcc7cd512cb61f1f038e53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/auth/signer.go
+++ b/auth/signer.go
@@ -65,31 +65,9 @@ func NewRequestSigner(endpoint, region string) *RequestSigner {
 	}
 }
 
-// UnsignedRequest builds an HTTP request to the upstream carrying NO credentials —
-// no Authorization, X-Amz-Date, or content hash — i.e. an anonymous request, exactly
-// like an unauthenticated client. Used to probe anonymous (public-read) access:
-// upstream serves it only if the object is publicly readable.
-func (s *RequestSigner) UnsignedRequest(ctx context.Context, method, path string) (*http.Request, error) {
-	baseURL, err := url.Parse(s.endpoint)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse endpoint: %w", err)
-	}
-
-	pathPart := path
-	queryPart := ""
-	if idx := strings.Index(path, "?"); idx != -1 {
-		pathPart = path[:idx]
-		queryPart = path[idx+1:]
-	}
-	baseURL.Path = pathPart
-	baseURL.RawQuery = queryPart
-
-	req, err := http.NewRequestWithContext(ctx, method, baseURL.String(), nil)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create request: %w", err)
-	}
-	req.Header.Set("Host", req.URL.Host)
-	return req, nil
+// Endpoint returns the upstream endpoint this signer targets.
+func (s *RequestSigner) Endpoint() string {
+	return s.endpoint
 }
 
 // SignRequest creates a new HTTP request signed for Tigris using streaming.
--- a/auth/signer.go
+++ b/auth/signer.go
@@ -65,6 +65,33 @@ func NewRequestSigner(endpoint, region string) *RequestSigner {
 	}
 }
 
+// UnsignedRequest builds an HTTP request to the upstream carrying NO credentials —
+// no Authorization, X-Amz-Date, or content hash — i.e. an anonymous request, exactly
+// like an unauthenticated client. Used to probe anonymous (public-read) access:
+// upstream serves it only if the object is publicly readable.
+func (s *RequestSigner) UnsignedRequest(ctx context.Context, method, path string) (*http.Request, error) {
+	baseURL, err := url.Parse(s.endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse endpoint: %w", err)
+	}
+
+	pathPart := path
+	queryPart := ""
+	if idx := strings.Index(path, "?"); idx != -1 {
+		pathPart = path[:idx]
+		queryPart = path[idx+1:]
+	}
+	baseURL.Path = pathPart
+	baseURL.RawQuery = queryPart
+
+	req, err := http.NewRequestWithContext(ctx, method, baseURL.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Host", req.URL.Host)
+	return req, nil
+}
+
 // SignRequest creates a new HTTP request signed for Tigris using streaming.
 // It accepts a pre-computed body hash (from X-Amz-Content-Sha256 header) to avoid
 // buffering the entire body in memory. The body is passed through as-is.
--- a/auth/signer_test.go
+++ b/auth/signer_test.go
@@ -10,31 +10,6 @@ import (
 	"testing"
 )
 
-func TestRequestSigner_UnsignedRequest(t *testing.T) {
-	signer := NewRequestSigner("https://s3.amazonaws.com", "us-east-1")
-
-	req, err := signer.UnsignedRequest(t.Context(), http.MethodGet, "/test-bucket/test-key")
-	if err != nil {
-		t.Fatalf("UnsignedRequest: %v", err)
-	}
-
-	if req.Method != http.MethodGet {
-		t.Errorf("method = %q, want GET", req.Method)
-	}
-	if got, want := req.URL.String(), "https://s3.amazonaws.com/test-bucket/test-key"; got != want {
-		t.Errorf("URL = %q, want %q", got, want)
-	}
-	// The whole point: no credentials of any kind are attached.
-	for _, h := range []string{"Authorization", "X-Amz-Date", "X-Amz-Content-Sha256"} {
-		if v := req.Header.Get(h); v != "" {
-			t.Errorf("unsigned request carries %s=%q, want empty", h, v)
-		}
-	}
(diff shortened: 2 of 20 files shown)


ovaistariq:
High-effort review surfaced four findings; addressed in the latest push.

**Three collapse into one root cause** — getting the warm's credentials by re-validating the write request (`ValidateAndGetCredentials(r)`):

1. **[Confirmed] Transparent-mode guard was dead / doc + test wrong.** `ValidateAndGetCredentials` always returns TAG's non-empty creds in transparent mode, so the `err/empty` guard never fired and `AuthResult` was ignored. The behavior (warming any upstream-accepted write) is fine, but the doc comment and the `SkippedWithoutCredentials` test claimed otherwise.
2. **[Confirmed] Signing-mode write-only client → 403 spam.** It returned the client's creds, so a write-only client's warm GET 403'd on every write, inflating `tag_background_fetches_failed_total` and logging a warning per write.
3. **[Confirmed, plausible] Slow-upload expiry.** Re-validating *after* the upload could trip the 15-min SigV4 max-request-age, silently skipping the warm for exactly the large objects.

**Fix:** replaced re-validation with a dedicated `forwarder.BackgroundFetchCredentials()` — TAG's own proxy read creds in transparent mode (`ok=true`), and `ok=false` in signing mode, where TAG has no read identity other than the client's. So warm-on-write is transparent-mode-only (documented), warms any upstream-accepted write including anonymous public-bucket writes, and does no per-write SigV4 re-validation. This closes 1/2/3 together.

4. **[Confirmed, lower] Dedup race.** If a read-path fetch of the *pre-write* object is already in flight, the warm coalesces into it and is dropped, so the new object isn't warmed — it's then populated on the next read. This is within the feature's documented best-effort/deduplicated semantics; I've documented it explicitly in `warmOnWrite` rather than adding a retry loop.

Tests updated (`SkippedWithoutCredentials` → `SkippedInSigningMode`); `make build`, `lint`, `test` green.

ovaistariq:
Both observations are accurate, but on analysis neither warrants a behavior change — both degrade safely to the cache-on-read baseline. I've documented the reasoning in code so they aren't later "fixed" into regressions.

**Newer warm is deduplicated (rapid same-key writes).** Correct: a warm coalesces into any in-flight fetch for the key (a read-path warm, or a prior write's warm) and is dropped, and the in-flight one is tombstone-blocked, so the latest object isn't warmed until a read triggers a fetch. But that read *is* the cache-on-read baseline — the object is never stale (a populate whose writeStartTime predates the latest write's invalidation tombstone is blocked), just not pre-warmed. It's the same best-effort coalescing already noted for the in-flight-read case, so I broadened that caveat. Notably the target workload (write-once keys, e.g. dated parquet) never hits this — it only bites rapid overwrites of one key, where the object is also least worth warming (it's about to be overwritten again). A coalesce-to-latest retry would fix it but adds a retry loop to machinery shared with the read path; not worth it for an opt-in best-effort feature.

**Warm skips public-read ACL.** This one is deliberate, and matching the read path (`hasNoAuthCredentials`) would be a **security regression**, not a fix. The read-path injection is sound because it fires on a successful anonymous *GET* — which proves the object is publicly **readable**. A successful anonymous *write* only proves the bucket is publicly **writable**; inferring public-read from it would let an anonymous reader be served a private object from a public-write/private-read bucket (`get_object.go:100` exists precisely to prevent serving non-public objects to anonymous readers). Leaving `publicACL=false` costs at most one anonymous miss after an anonymous write: that read's streaming populate (`get_object.go:361-364`) re-caches *with* public-read, correctly derived from the anonymous GET succeeding. So it self-heals safely rather than risking a leak.

`make build`, `lint`, `test` green.

ovaistariq:
**The failing `s3-compat-tests` check is a pre-existing upstream issue, not this PR.**

Only `test_s3.py::test_bucket_delete_nonempty` fails (both single-node and cluster), with "ClientError not raised" — the test does `PUT object` then `DELETE bucket` and expects a 409 `BucketNotEmpty`, but the delete is succeeding. That's a read-after-write consistency lag on bucket-emptiness in the upstream backend, not TAG.

Two independent proofs it isn't this PR:
1. **The diff is inert with the flag off.** The only non-test code is `warmOnWrite` (returns immediately when `cache.warm_on_write` is false — the default, and it is not set in the s3-test config) plus its config field/metric; the forwarder files net to zero change vs `main`. So this branch is behaviorally identical to `main` for every operation this test exercises, and the test never touches the warm path.
2. **It reproduces on `main`.** I re-ran the last green `main` run (commit `baca028`, which passed on 07-19) against today's upstream — it now fails the same `test_bucket_delete_nonempty`. Same unchanged code, now red → the change is upstream.

So this check will block any PR/merge until the upstream consistency issue is resolved (or the test is quarantined). It is not a blocker introduced by #113. Filing a separate issue to track the upstream regression + quarantine.

ovaistariq:
@greptile

ovaistariq:
Correction/clarification on the "newer warm is deduplicated" point above, and tightened the code comment to match:

The coalesced (dropped) warm is not the whole story — the **in-flight fetch it coalesces into is itself tombstone-blocked** when it predates this write (its `writeStartTime` is older than this write's invalidation tombstone), so it writes nothing either. So after a rapid same-key overwrite the key is left **absent, not stale**: the next read misses and inline-populates the current object via `setupCacheListener` (the cache-on-read baseline). A stale object can never be served here — the same tombstone that blocks the racing populate is the read-after-write guard (#97/#106).

So this is purely a "not pre-warmed" effectiveness gap for rapid overwrites of one key, never a stale-read correctness issue. Comment in `warmOnWrite` updated accordingly.

ovaistariq:
@greptile

PR #116, opened 2026-07-22
fix: never mark a cache entry public-read from a signed background fetch (P1)

ovaistariq:
Fixes a P1 security issue flagged in review on the read-path background fetch.

## Vulnerability

The read-path background-fetch callers (`get_object.go:301/742`, `revalidation.go:374`) passed `publicACL=hasNoAuthCredentials(r)` but `anonymous=false`. In transparent mode an anonymous request carries **TAG's proxy credentials**, so the background fetch ran **signed** — it could read a **private** object — and then, because `publicACL=true` for the anonymous request, marked the cached metadata **public-read**. A later anonymous request would then pass the `IsPublicRead` cache-authorization check and be served that private object from cache.

## Root cause & fix

`publicACL` and `anonymous` were separate parameters, but they must always be equal: **public-read may only ever be inferred from a successful UNSIGNED read** (which actually proves anonymous readability). Keeping them separate is what allowed the unsafe `signed fetch + mark public-read` combination.

Collapse them into a single `anonymous` flag threaded through `triggerBackgroundCacheFetch` / `fetchFullObjectToCache`, so the unsafe combination is impossible by construction:
- **Anonymous request** (`anonymous=true`): unsigned fetch → 200 caches public-read, 403 caches nothing.
- **Authenticated request** (`anonymous=false`): signed fetch, never marked public-read.

The read-path callers now pass `hasNoAuthCredentials(r)` (they previously passed a hardcoded `false` for the fetch mode).

## Tests

- Strengthened `TestSDK_PublicBucket_PrivateObjectDeniedAnonymously` with a repeat anonymous GET of the private object — directly catches a background fetch that cached a private object as public-read (a single GET wouldn't).
- The cold-owner warm unit test now uses an authenticated request so it exercises the signed-fetch path (an anonymous request now correctly routes to the unsigned fetch).
- Existing warm-on-write unit tests already cover the anonymous unsigned-fetch behavior (200 → public-read cached, 403 → not cached).

`make build`, `lint`, `test`, `test-integration` all green.

---

> [!NOTE]
> **High Risk**
> Changes cache authorization for background populates; incorrect behavior would leak private objects to anonymous clients, though the refactor makes the unsafe signed+public-read combination impossible by construction.
> 
> **Overview**
> Fixes a **P1 cache authorization bug** where background full-object warms could label entries **public-read** after a **signed** upstream read (using TAG credentials), letting later anonymous clients be served **private** objects from cache.
> 
> **`publicACL` and `anonymous` are merged into one `anonymous` flag** on `triggerBackgroundCacheFetch` / `fetchFullObjectToCache`. **Public-read is set only when `anonymous` is true** (unsigned fetch that proves anonymous readability). Signed warms never get public-read metadata.
> 
> Read-path triggers (`get_object`, revalidation) now pass **`hasNoAuthCredentials(r)`** instead of a hardcoded fetch mode plus a separate ACL flag. Warm-on-write uses the same single flag.
> 
> **Tests:** `TestSDK_PublicBucket_PrivateObjectDeniedAnonymously` polls repeat anonymous GETs to catch async background fetches that poison the cache; the cold-owner warm test uses an authenticated request so it exercises the signed-fetch path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47ca606bc7dbc497f139f855cad7e2f6b6a0c4ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/caching.go
+++ b/proxy/caching.go
@@ -317,10 +317,15 @@ func (s *Service) setupCacheListener(
 
 // fetchFullObjectToCache fetches the full object and caches it.
 // This makes a full-object request (no Range header) and streams directly to cache.
+// When anonymous is true the fetch is unsigned and, if it succeeds, the object is
+// cached as public-read. These are inseparable: public-read may be inferred ONLY
+// from a successful UNSIGNED read (which proves anonymous readability). A signed
+// fetch — which uses TAG's credentials and can read private objects — must never
+// mark an entry public-read, or a later anonymous request could be served a private
+// object from cache.
 func (s *Service) fetchFullObjectToCache(
 	ctx context.Context,
 	bucket, key, accessKey, secretKey string,
-	publicACL bool,
 	anonymous bool,
 	broadcaster *broadcast.Broadcaster,
 ) error {
@@ -409,10 +414,11 @@ func (s *Service) fetchFullObjectToCache(
 		return errCachePopulateDeclined
 	}
 
-	// If the original request was anonymous and succeeded, the object is publicly
-	// accessible. Tigris omits X-Amz-Acl for objects inheriting bucket-level access,
-	// so inject public-read to ensure the cached metadata allows anonymous reads.
-	if publicACL && resp.Header.Get("X-Amz-Acl") == "" {
+	// The unsigned fetch succeeded, so the object is anonymously readable. Tigris
+	// omits X-Amz-Acl for objects inheriting bucket-level access, so inject public-read
+	// to ensure the cached metadata allows anonymous reads. Gated on `anonymous`: a
+	// signed fetch can read private objects, so it must never mark an entry public.
+	if anonymous && resp.Header.Get("X-Amz-Acl") == "" {
 		resp.Header.Set("X-Amz-Acl", "public-read")
 	}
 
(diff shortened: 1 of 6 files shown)


ovaistariq:
Good catch — fixed in the latest push. The single `time.Sleep(500ms)` + one repeat GET was exactly the "sleep for synchronization" anti-pattern: a background fetch that caches the private object public-read *after* 500ms would land after the repeat GET already saw the upstream 403, so the test false-passes while the poisoned entry is still in flight.

Replaced it with a **poll** across a 3s window that fails on any `200`/`X-Cache: HIT`. The key property that makes this robust: with the bug, once the entry is cached public-read *every* subsequent anonymous GET leaks it, so any iteration in the window catches a 200/HIT; without the bug every iteration stays 403. The residual (a fetch outlasting the whole window) is inherent to a black-box end-to-end timing test, but 3s is well beyond a healthy background fetch of a tiny object.

Worth noting the deterministic guard for this fix lives in the proxy unit tests (anonymous background fetch is unsigned → 403 caches nothing, 200 caches public-read); this SDK test is the end-to-end defense-in-depth check against a real-Tigris ACL regression.

ovaistariq:
@greptile

Issue #123, opened 2026-07-24
perf: reduce cross-node gRPC data-plane cost (single-owner ring serves ~77% of reads remotely)


ovaistariq:
## Summary

~77% of TAG's burst CPU is spent moving cached bodies **between nodes over gRPC**, not serving clients. The single-owner consistent-hash ring means most range reads are served cross-node, roughly doubling network handling and adding gRPC serialization and buffer-pool churn on every byte. Improve **request locality** so cached bytes are served from the node that owns them.

## Evidence (production burst profile)

7-node cluster `parseable-tag`, real ~10.8-core burst (325 CPU-s sampled at 1077%), workload is 99.96% range-from-cache hits (`tag_range_from_cache_hits_total` ≈ `tag_cache_hits_total`), background full-object fetches only 1.5% of GETs:

| Share | What the CPU was doing |
|------:|------------------------|
| **64.8%** | Serving range reads to *other* nodes over gRPC (this node = key owner): `CacheService.Get → byteRangeReader.Read → pread` off disk segment files |
| **12.2%** | Pulling bodies *from* peers for own clients: `serveRangeFromCache → cache.GetRangeStream → embedded.GetRangeStream → getRemote (gRPC client)` |
|  2.4% | Background full-object fetch + remote put |
|  ~20% | Low-level data movement (memmove/cgo/syscall) + gRPC transport + GC |
|  **0.2%** | **Ideal: client request for a locally-owned key, served from local disk** |

Corroborating signal: bytes-in (14.98 GB) exceeds bytes-out (12.08 GB) on a read-only-serving node (inbound peer gRPC).

Routing path: `ocache/server/operations/get.go` — `Get()` calls `IsLocal(key)` (consistent-hash ring via coordinator); non-local keys go through `getRemote()` which opens a gRPC stream to `Route(key)` and `io.Copy`s the body back. With 7 nodes and uniform hashing, ~6/7 of reads are remote.

Flat CPU split (self-time, symbolized against a verified-complete binary): **83% in `Syscall6`** — cumulatively **~74% `pread`** (`byteRangeReader.Read → os.File.ReadAt → pread`, reading served bytes off ocache disk segment files) + **~6% socket `write`**. cgo/RocksDB is negligible (**0.6%** — an earlier "18% RocksDB cgo" figure came from a corrupted binary copy and is wrong; body reads are direct pread, not RocksDB). Copies ~4%, crypto minor. The range reader reads only the requested range (no gross over-read).

[...]

PR #124, opened 2026-07-24
feat: ocache v1.9.0 (range read-amplification fix) + serve-locality metric (#122)

### LINKED ISSUE #122: feat: expose local-vs-remote cache serve ratio metric (cross-node gRPC visibility)
## Summary

Add metrics that expose how many cache reads are served **locally** vs **pulled from a peer node over gRPC**, so cross-node data-plane cost is visible on dashboards without capturing pprof profiles.

## Motivation

Production CPU profiling of the 7-node cluster (`parseable-tag`) during a real ~10.8-core burst (2K req/s, ~99.96% range-from-cache hit ratio) attributed burst CPU as:

| Share | Path |
|------:|------|
| **64.8%** | Serving range reads to *other* nodes over gRPC (this node is the key owner) |
| **12.2%** | Pulling bodies *from* peers to serve this node's own clients |
|  0.2% | Ideal path: client request for a locally-owned key, served from local disk |
|  ~20% | Low-level data movement (memmove/cgo/syscall/TLS) + gRPC/TLS transport + GC |

So ~77% of burst CPU is the single-owner consistent-hash ring routing cached bodies through gRPC (`ocache/server/operations/get.go`: `Get → IsLocal? → getRemote`). None of this is visible in the current metrics — `tag_range_from_cache_hits_total` counts a hit whether it was served locally or pulled cross-node.

Diagnosing this required dumping raw pprof and hand-parsing embedded symbols (the pod's go1.24 runtime profiles mis-symbolize stdlib frames under both go1.24.2 and go1.26.1 `pprof`). A metric would make it a dashboard panel.

## Proposed metrics

- `tag_cache_serve_locality_total{locality="local"|"remote"}` — counter, incremented per range/object serve based on `IsLocal(key)` at serve time.
- Optionally `tag_cache_remote_get_bytes_total` and/or a per-owner label to surface **hot-key ownership imbalance** (in the burst, this node spent 5× more CPU serving peers than its own clients — one node owning a hot key becomes a cluster-wide hotspot).

## Acceptance

- Panel/PromQL for local-vs-remote serve ratio: `rate(tag_cache_serve_locality_total{locality="remote"}[5m]) / rate(tag_cache_serve_locality_total[5m])`.
- Documented in `docs/metrics.md`.
- Config-override test if any new toggle is introduced (none expected).

Blocks / informs #123.

ovaistariq:
## Summary

Two changes, one of them a significant performance fix:

1. **Range read-amplification fix (via ocache v1.9.0, tigrisdata/ocache#196)** — the important one. Ranged reads of segment/raw-file objects no longer read-and-discard the entire prefix before the requested range. This is the root cause of the ~74% `pread` burst CPU and the ~24 GiB page-cache "memory" found in #123. **Merging this PR is what deploys that fix to TAG.**
2. **`tag_cache_serve_locality_total{locality="local"|"remote"}` metric** — makes the cross-node cache data-plane cost visible on dashboards instead of only in CPU profiles. Closes #122.

Both arrive together because the metric needs `embedded.Client.IsLocal` (also new in ocache v1.9.0), so this PR bumps ocache and adds the metric that consumes it.

## The ocache v1.9.0 bump (all 8 submodules, v1.8.0 → v1.9.0)

- **Range read-amplification fix** (tigrisdata/ocache#196): ocache's `byteRangeReader` seeks to a range's start offset via an `io.Seeker` assertion, else it reads-and-discards the prefix. The segment/raw-file read wrappers embedded `io.Reader`, hiding the underlying `*io.SectionReader`'s `Seek` — so every ranged read with `start > 0` discarded `start` bytes (serving `[900 MB, +1 MB]` read and threw away 900 MB). Because the discarded bytes are page-cache-hot, it showed up as CPU (`copy_to_user`), not I/O wait. Fixed by embedding `io.ReadSeeker`.
- **`embedded.Client.IsLocal`** (tigrisdata/ocache#194): the capability this metric type-asserts for. Verified `*embedded.Client` now satisfies it, so the counter populates in cluster mode (it was intentionally inert against v1.8.0).

## Metric changes (TAG code in this PR)

- **metrics**: `tag_cache_serve_locality_total{locality}` counter + `RecordCacheServeLocality` helper and `LocalityLocal`/`LocalityRemote` constants.
- **cache**: record locality at all three body-read success points (`GetBodyStream` and both `GetRangeStream` branches) via the optional `IsLocal(key) bool` capability. When the client can't report ownership, the counter is left untouched rather than guessing — the ratio stays honest.
- **docs**: `docs/metrics.md` entry with the remote-ratio PromQL.

## Test

`go build ./...`, `go vet ./...`, gofmt clean on v1.9.0. `cache`/`metrics`/`proxy` suites green. `cache/locality_test.go` covers local/remote ownership, both range paths, misses (no record), and the no-capability graceful skip.

Acceptance from #122:
```promql
rate(tag_cache_serve_locality_total{locality="remote"}[5m]) /
rate(tag_cache_serve_locality_total[5m])
```

After deploy, re-profiling a burst should show `pread` collapse and working-set memory fall (verification tracked in #123).


pboros (dismissed):

ovaistariq:
@greptile

Issue #129, opened 2026-07-26
perf: link jemalloc to fix RocksDB glibc-malloc RSS fragmentation (memory keeps growing)


ovaistariq:
## Summary

Link the deployed TAG binary against **jemalloc** to fix RocksDB memory-fragmentation RSS growth. In production, pod working-set memory creeps steadily (~2 GiB → ~4.5 GiB over 2 days and still climbing) because RocksDB's allocation churn under **glibc malloc** scatters freed memory across per-thread arenas that are never returned to the OS.

## Production evidence (v1.14.0, parseable-tag cluster)

Diagnosed the "Memory (working set) keeps growing" graph:

- **Not a Go leak** — Go heap `inuse_space` = 105 MB, goroutines = 132 (stable).
- The growth is `anon` (RocksDB, cgo) **~1.8–2.25 GiB and climbing** + `slab` ~1.4 GiB. Page cache is separate and reclaimable.
- RocksDB is configured with bounded caches (1 GiB block cache with `cache_index_and_filter_blocks=true`, ~384 MB memtables) — a ~1.4 GiB logical budget — yet RSS far exceeds it and keeps growing for days. Bounded caches fill and plateau in minutes; multi-day creep is fragmentation.
- Allocator confirmation: **no jemalloc/tcmalloc linked** (0 `je_*` exports), **no `MALLOC_ARENA_MAX`** set, `VmData` 3.6 GB vs `RssAnon` 1.78 GB (the ~1.8 GB gap is reserved-but-unfreed glibc arena space — a fragmentation signature, not a live leak).
- Fragmentation driver: the container sees **64 CPUs** (glibc ignores the cgroup CPU limit) → up to **8×64 = 512 arenas**, with **114 threads** (RocksDB compaction/flush/block-cache) churning allocations.

This is the well-known RocksDB-on-glibc-malloc RSS-bloat pattern; RocksDB deployments almost universally run jemalloc.

## Constraint: static binary → link-time jemalloc

The released binaries are **statically linked** (glibc static rocksdb) and copied into an alpine image, so `LD_PRELOAD libjemalloc.so` will not work. jemalloc must be linked at build time.

## Proposed work

1. **jemalloc at link time** — add `-ljemalloc` to the RocksDB `CGO_LDFLAGS` in the `Makefile` (needs a static `libjemalloc.a` for each `ROCKSDB_PLATFORM`, alongside the existing `rocksdb-static` artifact from `ocache-releases.t3.storage.dev`).
[...]

PR #130, opened 2026-07-27
perf: link jemalloc into the Linux build to fix RocksDB RSS fragmentation

ovaistariq:
Implements the TAG-side of #129. The ocache side is done (jemalloc `rocksdb-static` artifact published + link recipe proven — ocache #198/#201), so no ocache release is needed; TAG consumes the versioned artifact and links it in its own build.

## Problem
The deployed binary is **statically-linked glibc**. RocksDB's allocation churn scatters across up to `8×nCPU` glibc arenas (container sees **64 CPUs → ~512 arenas**, 100+ threads) and retains freed memory instead of returning it to the OS, so pod working-set RSS creeps unbounded (~2 → ~4.5 GiB over 2 days on prod). Bounded RocksDB caches (~1.4 GiB) can't explain multi-day creep — it's fragmentation (analysis in #129).

## Changes (Linux deploy target only)
- **Artifact:** fetch `rocksdb-static-<ver>-jemalloc-Linux-<arch>.tar.gz` (RocksDB `-DWITH_JEMALLOC=ON` + bundled static `libjemalloc.a`). macOS dev builds keep the plain artifact.
- **Link flags:** `-ljemalloc -ldl` after `-lrocksdb` so jemalloc's `malloc` resolves RocksDB's allocations; `extldflags` gains `-Wl,--allow-multiple-definition` (a fully-static glibc binary also pulls glibc's `malloc.o`, which collides with jemalloc's — the flag lets jemalloc's first-seen definition win).
- **`verify-jemalloc` build gate:** `nm` proves `mallctl`/`malloc` are present (mallctl is jemalloc-exclusive), and an invalid-`MALLOC_CONF` runtime probe proves jemalloc is *active* at runtime (glibc ignores `MALLOC_CONF`). A silent no-op would look identical until prod RSS creeps, so this fails the build.

## Notes
- **macOS unaffected** — verified variable expansion: macOS fetches the plain artifact and skips verify.
- **Tests unchanged** — they must link `-ljemalloc` (to resolve `librocksdb.a`'s `je_*` refs) but link *dynamically* (no `-static`), so jemalloc interposes libc's malloc with no collision; only the static `build` needs `--allow-multiple-definition`.
- **CI validates it** — `release.yaml` runs `make build` on **native** x86_64 and arm64 runners, so the verify gate (including the runtime probe) runs per-arch.
- Couldn't fully exercise the Linux static+jemalloc link locally (macOS host); the recipe mirrors ocache #201, proven on both arches. CI is the real gate.
- **Final validation** is deploy-time: RSS before→after on the parseable-tag cluster (tracked in #129).

---

> [!NOTE]
> **Medium Risk**
> Changes the Linux binary’s memory allocator and static link recipe for all production builds; mistakes could fail CI/release builds or alter RSS behavior until validated in prod (#129).
> 
> **Overview**
> **Linux production builds** now pull the jemalloc RocksDB static tarball, link `-ljemalloc -ldl` after RocksDB, and use `-Wl,--allow-multiple-definition` on fully static links so jemalloc’s `malloc` wins over glibc’s. macOS keeps the plain artifact and skips jemalloc.
> 
> After `make build`, **`verify-jemalloc`** runs on Linux: `nm` checks for `mallctl`/`malloc`, and a `MALLOC_CONF=confirm_conf:true` runtime probe on `--version` proves jemalloc is actually active (not a silent link no-op).
> 
> **RocksDB artifact download** is safer and variant-aware: it records `.artifact`, re-downloads when the tarball name changes, and uses a `rocksdb-static.tmp/` staging dir (also gitignored). **CGO test targets** pass `TEST_LDFLAGS` for external linking so tests resolve jemalloc’s malloc without full static linking.
> 
> **CI** adds a `make build` step before tests so the jemalloc link and verify gate run on every test workflow. Deploy examples and K8s/native defaults bump pinned **`TAG_VERSION` from v1.12.0 to v1.14.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc71f54bb5ecb6f9ef0ece6f9dcf12837e0789aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,9 @@ jobs:
       - name: Run linting
         run: make lint-ci
 
+      - name: Build static binary (validates jemalloc link + verify gate)
+        run: make build
+
       - name: Run unit tests
         run: make test
 
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ coverage-integration.html
 
 # RocksDB static artifacts
 rocksdb-static/
+rocksdb-static.tmp/
 
 # Environment variables
 .env
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,11 @@ else
 	@echo "Verifying jemalloc is the linked, active allocator..."
 	@nm $(BINARY_NAME) | grep -qE ' T mallctl$$' || { echo "FAIL: jemalloc not linked (mallctl symbol absent)"; exit 1; }
 	@nm $(BINARY_NAME) | grep -qE ' T malloc$$'  || { echo "FAIL: malloc symbol not defined in binary"; exit 1; }
-	@MALLOC_CONF=verify:1 ./$(BINARY_NAME) --version 2>&1 | grep -q '<jemalloc>' || { echo "FAIL: jemalloc not active at runtime (glibc ignores MALLOC_CONF)"; exit 1; }
+	@# confirm_conf:true is jemalloc's documented option for confirming MALLOC_CONF is
+	@# read: it echoes each parsed conf pair prefixed with "<jemalloc>". glibc ignores
+	@# MALLOC_CONF entirely, so the marker is absent when jemalloc is not the allocator.
+	@# (Positive signal — does not rely on any key being unrecognized.)
+	@MALLOC_CONF=confirm_conf:true ./$(BINARY_NAME) --version 2>&1 | grep -q '<jemalloc>' || { echo "FAIL: jemalloc not active at runtime (glibc ignores MALLOC_CONF)"; exit 1; }
 	@echo "OK: jemalloc linked (mallctl + malloc) and active at runtime"
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,16 @@ UNAME_M := $(shell uname -m)
 # RocksDB artifacts use: Linux-x86_64, Linux-aarch64, macOS-arm64
 ifeq ($(UNAME_S),Darwin)
     ROCKSDB_PLATFORM := macOS-arm64
+    # macOS dev builds use the plain rocksdb-static artifact. jemalloc is only linked
+    # for the Linux deploy target (no macOS jemalloc artifact is published).
+    ROCKSDB_ARTIFACT_VARIANT :=
     BREW_PREFIX := $(shell brew --prefix 2>/dev/null || echo "/opt/homebrew")
 else
+    # Linux deploy target uses the jemalloc variant of the rocksdb-static artifact
+    # (RocksDB built -DWITH_JEMALLOC=ON + bundled static libjemalloc.a) so jemalloc is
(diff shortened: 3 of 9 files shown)


efirs (dismissed):

ovaistariq:
@greptile

ovaistariq:
Good catch, @cursor — valid. The `rocksdb-static` target keyed the skip only on `include/` existing, so a previously-fetched **plain** tree (or an older `ROCKSDB_VERSION`) would be reused and `-ljemalloc` couldn't resolve without a manual `make rocksdb-static-clean`. Fixed by recording the fetched artifact name in a `.artifact` marker and re-fetching whenever it differs from the wanted `ROCKSDB_STATIC_ARTIFACT` — covers both the plain→jemalloc switch and future version bumps. (CI runners fetch fresh, so this was a local-dev / cached-workspace trap.)

ovaistariq:
Both fixed (7cae6e1 → new commit):

- **Destructive re-fetch masks failures** (Medium) — valid. Now download+extract into `rocksdb-static.tmp` and only `rm -rf`+`mv` into place on success, with the whole sequence `&&`-chained. A failed fetch exits non-zero (surfaces the error instead of a confusing link failure) and leaves the existing tree intact. Verified: simulated download failure → exit 1, existing tree preserved.
- **Missing help for new target** (Low) — added `verify-jemalloc` to `make help` with a usage example, per CLAUDE.md.

The earlier `Makefile:61` stale-artifact finding was already addressed by the marker check in 7cae6e1.

ovaistariq:
@greptile

ovaistariq:
Fixed — valid. `rocksdb-static-clean` now also removes `rocksdb-static.tmp` (so `clean-all`, which chains it, covers it too), and `.gitignore` now ignores `rocksdb-static.tmp/`. Verified `git check-ignore` matches the staging dir.

ovaistariq:
Improved the probe, though the severity is overstated by one key fact: **the gate was not failing**. The first CI run (commit `77f5459`) printed `OK: jemalloc linked (mallctl + malloc) and active at runtime` — i.e. the probe found `<jemalloc>` and passed. So in the shipped jemalloc 5.3.0, `verify:1` is treated as an *unrecognized* conf pair and does emit the `<jemalloc>` warning; it is not silently accepted, and Linux builds were not failing.

That said, the underlying fragility is real — relying on a key being unrecognized is brittle (a future jemalloc could add `opt.verify`, or change unknown-key handling). Switched the probe to `MALLOC_CONF=confirm_conf:true`, jemalloc's documented option for confirming `MALLOC_CONF` is read (it echoes each parsed conf pair prefixed with `<jemalloc>`). That's a positive signal that doesn't depend on any key being invalid; glibc still ignores `MALLOC_CONF`, so the negative case is unchanged. CI will re-confirm on this commit.

PR #135, opened 2026-07-29
feat: write-through cache tee for single PutObject (#134)

### LINKED ISSUE #134: feat: write-through cache tee for single-object PutObject (avoid warm-on-write read-back)
## Summary

For single-object `PutObject`, populate the cache by **teeing the request body as it is forwarded upstream**, instead of triggering a separate `warm_on_write` background GET that reads the just-written object back from upstream. TAG already has the assembled body in hand on the PUT path, so the read-back is pure waste.

Multipart completions and `CopyObject` keep the existing `warm_on_write` GET (TAG never sees their assembled body).

## Motivation

Today, on a successful write TAG caches the object via `warmOnWrite` → `triggerBackgroundCacheFetch`, which issues a **full-object GET to upstream** and streams the result into the cache:

- `HandlePutObject` calls `s.warmOnWrite(r, bucket, key)` — `proxy/service.go:416`
- `warmOnWrite` → `triggerBackgroundCacheFetch` → `fetchFullObjectToCache` (a background upstream GET) — `proxy/service.go:636`, `proxy/caching.go`

This means every written object incurs **write-up + read-back-down** across the upstream link. For a single `PutObject`, TAG already streamed the full body to upstream and can cache it directly. The read-back buys nothing and costs:

1. **Upstream egress/bandwidth**: every written object is re-fetched in full. For a write-heavy or TAG-only-cached deployment this doubles write-path upstream traffic.
2. **A write→warm latency window**: between the write completing and the async warm finishing, reads miss (cold), and the warm can lose priority via the documented coalescing edge (a read-miss fetch already in flight for the key). See the caveat comment on `warmOnWrite` (`proxy/service.go:603-635`).
3. **Populate-budget contention**: the read-back competes for the cache-populate byte budget with genuine read-miss populates.

A write-through tee makes the object **hot the instant the write succeeds**, with no read-back and no warm race.

### Concrete driver

For the Parseable workload (which relies on TAG as its sole cache), data files are written as **single `PutObject` of immutable Parquet, predominantly < 25 MB**, and are queried shortly after. Teeing these on write eliminates a full read-back per file and removes the cold-read window before the first query.

## Current behavior (relevant code)

- `signingForwarder.Forward` decodes any AWS chunked encoding and holds the assembled body: `body, bodyHash, contentLength, chunked := decodeChunkedIfNeeded(r)` — `proxy/forwarder_signing.go:28`. The forward then streams `body` upstream.
- Response capture already uses a tee: `capture.Body, readErr = io.ReadAll(io.TeeReader(resp.Body, w))` — `proxy/forwarder.go:258`.
- Tombstone-aware cache write exists: `s.cache.PutWithMetaStreamTombstoneAware(ctx, bucket, key, meta, reader, ttl, writeStartTime)` — `proxy/caching.go:215`.
- Cacheability gate exists: `meta.IsCacheable(s.config.Cache.SizeThreshold)` / `isWithinSizeThreshold` — `proxy/caching.go:185,562`.

So all the primitives are already present; this issue wires them together on the PUT path.

## Proposed change

On `HandlePutObject`, when the object is cacheable (see gating below):

1. **Tee the assembled request body** (`decodeChunkedIfNeeded` output) into a bounded in-memory buffer while streaming to upstream — mirroring the existing response `TeeReader` pattern.
2. **Stamp `writeStartTime` before forwarding** (as `warmOnWrite` does) so a concurrent invalidation that is provably newer blocks the cache write.
3. **After upstream confirms success** (2xx, and not a 200-with-`<Error>` body — reuse `s3WriteSucceeded`), build the cache `meta` from the upstream **response** headers (crucially the returned `ETag`, since body keys are ETag-versioned — see commit `9e396de`) and write `meta` + the teed `body` via `PutWithMetaStreamTombstoneAware`.
4. **On any write failure or non-cacheable case, fall back to today's behavior** (no cache write; optionally still `warmOnWrite`).

Skip the `warm_on_write` GET entirely for objects handled by the tee.

## Scope / gating

**Tee applies to:**
- Single `PutObject` (`HandlePutObject`) …
- … with a known/decoded body **≤ `Cache.SizeThreshold`** (bounds the buffered memory; general S3 single PUT can be up to 5 GB, so a size gate is mandatory).
- Successful writes only (2xx, non-error body).

**Keep existing `warm_on_write` GET for (non-goals of this issue):**
- `CompleteMultipartUpload` (`proxy/service.go:712`) — TAG never sees the assembled body; see the existing comment at `service.go:710`.
- `CopyObject` (`proxy/service.go:560`) — server-side copy; TAG has no body.
- Single PUTs larger than `SizeThreshold` — fall back to current behavior.

## Design considerations

- **Memory bound**: buffer only up to `SizeThreshold`; account the tee-write against the existing cache-populate byte budget / slot (reuse `acquireCacheSlot` with `priorityWarmWrite`) so concurrent PUTs can't exhaust memory. Under budget saturation, fall back to `warm_on_write` (or skip caching) rather than block the client write.
- **ETag ordering (v1 buffer vs v2 decoupling)**: the ETag-versioned body key (`body|bucket|key|<etag>`, commit `9e396de`) is only known after the PUT response, so **v1 buffers the body (≤ `SizeThreshold`) and writes it post-response** — simple, correct, and fully sufficient for the small-object driver. A future **v2** could remove buffering *and* the size cap by decoupling the body key from the ETag — store the body under a TAG-generated token in `meta.BodyRef` (known before the response, so the body streams straight through) and keep the ETag only in `meta`. v2 is a cache-key-schema change and **must ship with a backward-compatible read** (resolve `meta.BodyRef` if present, else fall back to `meta.ETag`) so it does not invalidate existing cached entries and cold-start the live cache. Out of scope for v1.
- **Read-after-write / tombstone**: use `PutWithMetaStreamTombstoneAware` with the pre-forward `writeStartTime`, identical to the warm path, so a racing DELETE/overwrite is never resurrected.
- **Chunked uploads**: tee the *decoded* body from `decodeChunkedIfNeeded`, not the raw chunked stream.
- **Client latency**: the client PUT must not wait on the cache write — do the cache write after the client response is sent (or concurrently), best-effort, exactly like `warm_on_write` is best-effort today.

## Memory accounting (v1)

The tee-write is a cache-populate operation, so it must reuse the **existing cache-populate byte budget** (`Cache.MaxPopulateMemoryBytes` and the priority reservation from #132) — **not** a new memory pool:

- **Admission via `acquireCacheSlot(ctx, weight, priorityWarmWrite)`** — the same gate `fetchFullObjectToCache` uses. This bounds the tee under the single unified populate memory limit, gives it warm-write priority, and provides graceful backpressure.
- **Weight = actual buffered size** (the PUT `Content-Length`, capped at `SizeThreshold`) — **not** the streaming `perPopulateBufferBytes` weight. Streaming populates only hold a bounded channel buffer (~80 MB, size-independent) because they stream chunks; the buffered tee holds the **whole body in RAM** until the response, so the reservation must reflect the true bytes held. Reserving the small streaming weight would under-count the footprint and risk OOM under concurrent large PUTs.
- **Aggregate bound is automatic**: the budget caps total buffered-tee memory (e.g. 8 GiB budget ÷ 25 MiB ≈ ~320 concurrent tees; at Parseable's ~3 writes/s, contention is effectively nil).
- **Reuse the pooled buffers**: draw the buffer from the existing ocache buffer pool (`bufferpool` / `sync.Pool`) rather than a fresh allocation, to avoid per-write GC churn.
- **Reservation lifetime = buffer lifetime**: acquire before buffering, release after the cache write completes (or on fallback).
- **Non-blocking fallback**: if the budget can't admit the weight, do **not** block the client PUT — forward it uncached and let a later read-miss populate handle it (or skip). The client write path must never wait on cache admission.

## Benefits

- Eliminates one full-object **upstream read-back per single-PUT object** (roughly halves write-path upstream traffic for those objects).
- Object is **hot immediately** on write success — no cold-read window, no warm-coalescing race for single PUTs.
- Less cache-populate-budget contention (no separate GET populate for these writes).

## Acceptance criteria

- [ ] A single `PutObject` ≤ `SizeThreshold` results in a cache entry (`meta` + ETag-versioned body) with **no** subsequent upstream GET for that key.
- [ ] The very next GET/range-GET for that key is a cache hit (no cold miss, no background fetch).
- [ ] Multipart completion and `CopyObject` still populate via `warm_on_write` (unchanged).
- [ ] Single PUT > `SizeThreshold` falls back to current behavior; client write latency is unaffected in all cases.
- [ ] Tee buffer memory is charged to the existing populate byte budget (no new pool); when the budget is saturated the PUT is forwarded uncached with no added client latency.
- [ ] A DELETE/overwrite racing the tee-write is not resurrected (tombstone test), matching existing warm-on-write read-after-write guarantees.
- [ ] Metric visibility: tee-populated writes are distinguishable from read-back warms (e.g. a `source` label or a dedicated counter).

## Notes

- Complements, does not replace, `warm_on_write` — the two cover disjoint cases (assembled-body-visible vs not).
- Related: the cache-populate priority/budget work (#132) and ETag-versioned body keys (`9e396de`).

ovaistariq:
Closes #134.

## Summary

On an authenticated single `PutObject` within `cache.size_threshold`, populate the cache by **teeing the request body as it's forwarded upstream** instead of the read-back warm-on-write GET. In signing mode TAG already has the assembled body (after chunked decode), so the read-back is pure waste — this eliminates one full-object upstream fetch per written object and makes the object hot without a cold-read window.

Object **metadata** is sourced from a **HEAD** (which returns exactly the headers a GET would, minus the body), so the cached entry is authoritative and can't diverge from an origin GET. Net cost vs warm-on-write: a bodyless HEAD instead of a full-body GET — the bandwidth win is preserved.

## Design

- **Tee** the decoded body into a capped buffer while forwarding (`signingForwarder.ForwardTeeingBody`, exposed as an optional `bodyTeeingForwarder` capability). The capped buffer never errors, so it's a safe `io.TeeReader` sink.
- **HEAD for authoritative metadata** in the async populate path (reuses `DoConditionalHeadRequest` with empty conditions), build meta via `MetaFromHTTPHeaders`, write it with the teed body — no full-object read-back.
- **ETag-consistency guard**: cache only if the HEAD's ETag matches the PUT-response ETag. A concurrent overwrite returns a different ETag whose body isn't what we hold; caching it would pair mismatched meta/body (the torn-pair hazard ETag-versioned bodies prevent).
- **Tombstone/read-after-write**: `writeStartTime` is stamped **before** the HEAD (mirroring warm-on-write) so the whole HEAD→write window is guarded — the ETag guard catches pre-HEAD overwrites, the tombstone catches later ones, and our own post-forward invalidation doesn't block us.
- **Memory accounting**: reuses the existing cache-populate byte budget via a non-blocking `acquireCacheSlot` (weight = object size) — never blocks the client PUT, no new memory pool; on budget saturation it falls back to warm-on-write. Slot released when the async work completes.
- **Credentials** validated once by `ForwardTeeingBody` and threaded through; the async path never touches `r`.

## Scope / fallbacks

The tee applies to **authenticated single PUTs ≤ `size_threshold` in signing mode, when `warm_on_write` is enabled**. Everything else falls back to warm-on-write (unchanged): multipart completion and `CopyObject` (TAG can't see the assembled body), over-threshold, anonymous writes (keep warm's public-read probe), budget-saturated, and any case where the HEAD can't confirm the just-written version (HEAD failure / ETag mismatch / size disagreement).

## Metric

`tag_cache_write_through_total` — objects cached via the tee (no read-back). Compare with `tag_warm_on_write_triggered_total` (the read-back fallback).

## Tests

`proxy/write_through_test.go`: cache-from-HEAD (meta comes from the HEAD), ETag-mismatch and HEAD-failure both fall back to warm, a tombstone-during-HEAD blocks the stale write, warm-on-write-off disables the tee, and above-threshold falls back to warm. Full `proxy`/`metrics` suite passes with `-race`.

## Follow-up (not in this PR)

**v2**: stream the body into ocache under a token with `meta.BodyRef` (backward-compatible `BodyRef`-else-`ETag` read) to drop the buffer and its size cap for large objects. Tracked in #134.

---

> [!NOTE]
> **Medium Risk**
> Changes PutObject cache population, concurrency, and memory accounting on the write hot path; mistakes could serve stale or mismatched cached data, though guards and extensive tests mitigate this.
> 
> **Overview**
> When **`warm_on_write`** is on, eligible **authenticated single `PutObject`** requests in **signing mode** can populate the cache by **teeing the decoded body** while forwarding upstream, using a **HEAD** for authoritative metadata instead of a full read-back GET.
> 
> **`write_through_max_size`** (default 25 MiB, env `TAG_CACHE_WRITE_THROUGH_MAX_SIZE`) caps in-memory buffering; larger objects still use streaming warm-on-write. **`PutWithMetaStreamTombstoneAware`** now returns **`(wrote bool, err error)`** so callers can tell a skipped write (tombstone, no ETag) from a real cache entry and fall back to warm. **`tag_cache_write_through_total`** tracks successful tee populates.
> 
> Safety checks include ETag match with the PUT response, tombstone-aware writes, populate budget via non-blocking slot acquire, and fallbacks to warm-on-write on HEAD failure, mismatch, `no-store`, or budget pressure. Transparent proxy, anonymous writes, multipart, and over-cap objects are unchanged.
> 
> Docs and **`proxy/write_through_test.go`** cover the behavior; Python s3-tests teardown gets **`Tigris-Force-Delete`** for bucket cleanup only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b40e12a4136332061e18c037b53b9b076b7ea6ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -150,6 +150,11 @@ func (c *Cache) PutWithMeta(ctx context.Context, bucket, key string, meta *Cache
 // the orphaned body is cleaned up. Checking after body stream (rather than
 // before) closes the TOCTOU window where an invalidation during streaming
 // could be missed, causing resurrected metadata without a body.
+//
+// Returns wrote=true only when the metadata was actually written (the entry is now
+// visible). It is false when the write was skipped without error — the object was not
+// cacheable (no ETag) or a newer tombstone superseded it — so callers can distinguish a
+// no-op from a real write (e.g. for metrics or a fallback).
 func (c *Cache) PutWithMetaStreamTombstoneAware(
 	ctx context.Context,
 	bucket, key string,
@@ -157,9 +162,9 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 	body io.Reader,
 	ttl int,
 	writeStartTime int64, // Unix nano timestamp when write started
-) error {
+) (wrote bool, err error) {
 	if !c.IsEnabled() {
-		return nil
+		return false, nil
 	}
 
 	// Objects without an ETag are not cached: they would share a single
@@ -170,7 +175,7 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 	// Drain the body first so the producer side of the pipe never blocks.
 	if meta.ETag == "" {
 		_, _ = io.Copy(io.Discard, body)
-		return nil
+		return false, nil
 	}
 
 	if ttl == 0 {
@@ -184,13 +189,13 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 	metaBytes, err := meta.Encode()
 	if err != nil {
 		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).Msg("Cache meta encode error")
-		return err
+		return false, err
 	}
 
 	// Stream body to cache
 	if err := c.client.PutStream(ctx, bodyKey, body, int64(ttl)); err != nil {
 		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).Msg("Cache body put error")
-		return err
+		return false, err
 	}
 
 	// Check tombstone AFTER body stream, right before meta write.
@@ -209,7 +214,7 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 		// a reader streaming this exact body key, and deleting it would truncate
 		// that reader. Without a visible meta entry the orphaned body is unreachable
 		// and harmless until it expires.
-		return nil
+		return false, nil
 	}
 
 	// Write metadata AFTER body (makes entry visible)
@@ -217,7 +222,7 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 		log.Debug().Err(err).Str("bucket", bucket).Str("key", key).Msg("Cache meta put error")
 		// Same rationale as the tombstone branch: leave the versioned body to TTL
 		// rather than risk truncating a concurrent same-version reader.
-		return err
+		return false, err
 	}
 
 	log.Debug().
@@ -226,7 +231,7 @@ func (c *Cache) PutWithMetaStreamTombstoneAware(
 		Int("ttl", ttl).
 		Int("meta_size", len(metaBytes)).
 		Msg("Cached object with metadata (streamed, tombstone-aware)")
-	return nil
+	return true, nil
 }
 
 // GetMeta retrieves only object metadata from cache (no body).
--- a/cache/etag_versioning_test.go
+++ b/cache/etag_versioning_test.go
@@ -128,7 +128,7 @@ func TestETagVersionedBody_EmptyETagNotCachedViaStream(t *testing.T) {
 
 	body := bytes.NewReader([]byte("body-bytes"))
 	meta := &CachedObjectMeta{Bucket: bucket, Key: key, ETag: "", ContentLength: 10, StatusCode: 200}
-	if err := c.PutWithMetaStreamTombstoneAware(ctx, bucket, key, meta, body, 60, 1); err != nil {
+	if _, err := c.PutWithMetaStreamTombstoneAware(ctx, bucket, key, meta, body, 60, 1); err != nil {
 		t.Fatalf("PutWithMetaStreamTombstoneAware: %v", err)
 	}
 	if _, found, _ := c.GetMeta(ctx, bucket, key); found {
--- a/config/config.go
+++ b/config/config.go
@@ -38,6 +38,11 @@ const (
 	// DefaultCacheSizeThreshold is the max object size to cache (1GB).
 	DefaultCacheSizeThreshold = 1024 * 1024 * 1024
 
+	// DefaultCacheWriteThroughMaxSize caps the object size eligible for the write-through
+	// tee (16 MiB). It is far smaller than DefaultCacheSizeThreshold because the tee buffers
+	// the whole object in memory while forwarding the PUT, whereas the read/warm paths stream.
+	DefaultCacheWriteThroughMaxSize = 16 * 1024 * 1024
+
 	// DefaultCacheDiskPath is the default disk path for embedded cache storage.
 	DefaultCacheDiskPath = "/var/cache/tag"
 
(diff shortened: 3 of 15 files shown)


ovaistariq:
@greptile

ovaistariq:
@greptile

ovaistariq:
@greptile

PR #148, opened 2026-08-10
perf(proxy): avoid cache populate completion wait

perfloop-agent:
## Summary

- Return a cacheable full-GET fetcher after upstream EOF instead of blocking on detached cache population.
- Let broadcaster completion close cache listeners while preserving immediate cache-write error observation and late-error logging.
- Prefer ready broadcast headers over concurrent request cancellation so a completed response can still publish its cache entry, with regression coverage for the cancellation boundary, cache hit, late error, and slot release.

## Behavior

Cache publication remains asynchronous and subject to the existing admission and write-timeout limits. A follow-up request can still be a miss until the detached populate finishes.

## Testing

- `make lint-ci`
- `make test-proxy`
- `go test -race -timeout 120s ./proxy/...`
- `go vet ./proxy/...`

**Workload:** `cacheable-full-get-miss-handler-completion`

| Metric | Before | After | Change |
| --- | --- | --- | --- |
| `ns/op` | 101216495 | 572750 | 99.4% lower |
All 4 declared correctness checks passed.

---

Authored and verified by [Perfloop](https://app.perfloop.ai): every claim above was co-measured on both trees and independently re-verified before submission — the full record is public: [case_mtzbwvz133](https://app.perfloop.ai/t/oss/case_mtzbwvz133). Replies from this account are human-approved, and a human operator is accountable for this contribution.

---

> [!NOTE]
> **Medium Risk**
> Changes timing and cancellation behavior on the hot GET/cache-populate path; correctness relies on detached listeners and late-error handling, with substantial new tests but real production concurrency edge cases.
> 
> **Overview**
> Cacheable full-GET misses now finish the upstream streaming path as soon as the body is read, instead of waiting up to **100ms** (or longer) for detached cache population.
> 
> In **`streamFromUpstream`**, the handler no longer blocks on `cacheErrCh`; it logs an immediate cache-write error if one is already available, otherwise spawns a goroutine to log late failures. **`fetchAndBroadcast`** can call **`broadcaster.Complete`** after upstream EOF so listeners close promptly while populate continues in the background.
> 
> **`Listener.WaitForHeaders`** re-checks `headerCh` when `ctx` is already canceled, so a cache listener scheduled after the HTTP handler returns still sees ready headers and can publish the entry instead of losing the race to cancellation.
> 
> Regression coverage adds benchmarks and tests for handler completion timing, a follow-up cache hit on a separate connection, the header-vs-cancel boundary, late cache-write logging, and populate slot release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80112907fdebf08426ace9cc2494b742c02eceb7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/broadcast/broadcaster.go
+++ b/proxy/broadcast/broadcaster.go
@@ -94,13 +94,24 @@ type Listener struct {
 	disconnected bool          // True if disconnected due to slow consumption
 }
 
-// WaitForHeaders blocks until headers are available.
+// WaitForHeaders blocks until headers are available. If headers and context
+// cancellation become ready together, headers win. A cache listener can be
+// scheduled after its request handler returns, when both are already ready; it
+// must still consume the completed response and publish the cache entry.
 func (l *Listener) WaitForHeaders(ctx context.Context) (int, http.Header, error) {
 	select {
-	case <-ctx.Done():
-		return 0, nil, ctx.Err()
 	case <-l.headerCh:
 		return l.status, l.headers, nil
+	case <-ctx.Done():
+		// A select chooses randomly when both channels are ready. Re-check the
+		// headers so a completed response wins over concurrent cancellation, while
+		// still honoring cancellation when headers have not arrived.
+		select {
+		case <-l.headerCh:
+			return l.status, l.headers, nil
+		default:
+			return 0, nil, ctx.Err()
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


perfloop-agent:
I have read the CLA Document and I hereby sign the CLA

ovaistariq:
Thanks for the contribution. This looks good.

ovaistariq (approved):

PR #174, opened 2026-08-20
feat(cache): parquet footer prefetch and prefetch precision metrics

ovaistariq:
Tier 1 of the block-prefetching work: when a read reaches a parquet object's tail, fetch the metadata blocks the reader is about to need.

A parquet reader can't read data before it reads the file's metadata, and that metadata sits at the **end** of the object (4-byte LE length + `PAR1` magic in the last 8 bytes). So a read that reaches the trailer is a reliable signal the reader is opening the file.

## This is not a marginal case — measured on production

The triggering read caches only the **tail block**, and that's a *partial* block: `ContentLength mod block_size`, averaging half a block. So the prefetch does work whenever `footer + 8` exceeds that remainder — a lower bar than "footer exceeds `block_size`".

I sampled 26 objects from a production bucket across two days, reading each trailer directly:

| Object size | Footer | Blocks metadata spans (1 MiB) |
| --- | --- | --- |
| 394 MB | **4.69 MB** | ~5 |
| 342 MB | **4.31 MB** | ~5 |
| 269 MB | **3.53 MB** | ~4 |
| 244 MB | **3.01 MB** | ~3 |
| 133 MB | 1.62 MB | ~2 |
| 0.2–1.9 MB | 19–52 KB | 0 (fits) |

Footers are a consistent **~1.25% of object size** — the schema is 76 KB wide, with per-row-group column statistics on top. **The prefetch fires on 18 of 26 objects.** The exceptions are the small sub-2 MB files.

So on a large object today, a reader's first open finds 3–5 metadata blocks absent.

## How it's bounded

- Fires only on a **completed** serve that covered the trailer.
- Triggered from **both** serve paths. A trailer probe is a few bytes, so it lands in `serveAssembledRange`, not `streamBlockRange` — wiring only the latter meant it never fired (caught in review, 99d9b2a).
- The footer length comes from the **bytes just served**, not a cache read: the assembled path returns a freshly fetched tail from an in-memory lease while its cache write is still detached, so re-reading it would miss on exactly the cold open this targets (caught in review, 9632b4a).
- Range computed from the length the file declares, validated against object size. Bad magic or an impossible length → no prefetch.
- Capped at 32 blocks; skips blocks already cached.
- Coalesced per `bucket/key/etag`, so a hot object spawns one goroutine, not one per request.
- Uses `fetchOneBlock`, so it takes the populate budget **non-blocking** and is shed rather than queued under contention.

Off by default (`cache.parquet_optimization` / `TAG_CACHE_PARQUET_OPTIMIZATION`): it reads 8 bytes of object content, which is format-specific behaviour an operator should opt into.

## Metrics

| Metric | Purpose |
| --- | --- |
| `tag_cache_parquet_footer_bytes` | Footer-size distribution, recorded for every parquet object whose trailer is read — including non-prefetched ones, so it's unbiased. Now has a known baseline (~1.25% of object size) to detect **drift** against if the schema changes. |
| `tag_cache_block_prefetched_total` / `_prefetch_used_total` | Prefetch precision. Reusable by future triggers via the `trigger` label. |

Attribution counts **blocks, not reads**, clears atomically (concurrent serves can't double-count), and expires — so the ratio is a **floor**, never an inflated claim.

## Scope

The workload's objects are minute-sharded, so widening a query time range crosses into *different* objects rather than reading further within one. Same-object readahead (Tier 2) and cross-object prefetch (Tier 3) are deliberately out; Tier 3 would need LIST plumbing that doesn't exist. Block hit ratio is already 98.4%, so the footer is where a same-object optimization has a provable mechanism.

## Testing

`proxy/block_cache_parquet_test.go`: trigger conditions, trailer parsing (valid / bad magic / impossible length), coalescing, single-count attribution, plus three end-to-end tests — correct block range prefetched; a real trailer probe through `serveRangeFromBlockCache`; and a **cold open with nothing seeded**, asserting the served trailer is used when the tail isn't cached yet.

Both regression tests were verified to **fail without their fix** rather than assumed to pass — the first reported no upstream ranges at all, the second an empty range list.

`TestParquetOptimization_OverrideByEnv` per the repo's env-override convention, including that an unrecognized value leaves this speculative feature **off**.

## Rollout

Ship disabled, enable on one ORD node, and watch `prefetch_used / prefetched` alongside block-miss rate on first opens. The footer-size question that would normally gate this is already answered above.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -199,6 +199,17 @@ type CacheConfig struct {
 	// "lru" (default) or "fifo" (oldest-written first). Only takes effect when
 	// MaxDiskUsageBytes > 0 — with no disk cap nothing is evicted regardless.
 	EvictionPolicy string `yaml:"eviction_policy"`
+	// ParquetOptimization enables format-aware prefetching for objects whose key
+	// ends in ".parquet". A parquet reader always reads the file footer before any
+	// data, and the footer is at the end of the object, so a read that touches the
+	// object's tail is a reliable signal that the whole metadata region is about to
+	// be read. When the metadata spans more than the tail block, TAG fetches the
+	// remaining metadata blocks in the background instead of letting the reader
+	// discover them as serial misses. Off by default: it reads 8 bytes of object
+	// content to size the footer, which is format-specific behavior an operator
+	// should opt into.
+	ParquetOptimization bool `yaml:"parquet_optimization"`
+
 	// CompactionBytesPerSecond bounds the shared source-read budget for ocache's
 	// background file compaction and segment recompaction (bytes/second). It
 	// protects serving reads on throughput-capped volumes: unthrottled
@@ -573,6 +584,10 @@ func applyEnvOverrides(cfg *Config) {
 		// Override startup recovery worker count from environment
 		if workers, ok := envInt("TAG_CACHE_RECOVERY_WORKERS"); ok && workers > 0 {
 			cfg.Cache.RecoveryWorkers = workers
+		}
+		// Enable parquet-aware footer prefetching from environment.
+		if val := strings.ToLower(strings.TrimSpace(os.Getenv("TAG_CACHE_PARQUET_OPTIMIZATION"))); val != "" {
+			cfg.Cache.ParquetOptimization = val == "true" || val == "1"
 		}
 		// Override the compaction source-read budget from environment
 		// (bytes/second). An explicit 0 disables the throttle even when YAML
--- a/config/config.go
+++ b/config/config.go
@@ -203,11 +203,13 @@ type CacheConfig struct {
 	// ends in ".parquet". A parquet reader always reads the file footer before any
 	// data, and the footer is at the end of the object, so a read that touches the
 	// object's tail is a reliable signal that the whole metadata region is about to
-	// be read. When the metadata spans more than the tail block, TAG fetches the
-	// remaining metadata blocks in the background instead of letting the reader
-	// discover them as serial misses. Off by default: it reads 8 bytes of object
-	// content to size the footer, which is format-specific behavior an operator
-	// should opt into.
+	// be read. When the metadata spans more than the (partial) tail block, TAG
+	// fetches the remaining metadata blocks in the background instead of letting the
+	// reader discover them as misses. Measured on production parseable data, footers
+	// run ~1.25% of object size, so a 300 MB object carries ~3.5 MB of metadata --
+	// several blocks at a 1 MiB block_size. Off by default: it reads 8 bytes of
+	// object content to size the footer, which is format-specific behavior an
+	// operator should opt into.
 	ParquetOptimization bool `yaml:"parquet_optimization"`
 
 	// CompactionBytesPerSecond bounds the shared source-read budget for ocache's
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1240,3 +1240,34 @@ func TestCompactionBPS_OverrideByEnv(t *testing.T) {
 		})
 	}
 }
+
+// TestParquetOptimization_OverrideByEnv verifies TAG_CACHE_PARQUET_OPTIMIZATION
+// plumbs the parquet footer prefetch into the cache config. The optimization is
+// speculative work, so an unrecognized value must leave it off rather than
+// guessing the operator meant to enable it.
+func TestParquetOptimization_OverrideByEnv(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		yaml bool // value already set (as if from YAML) before the env override
+		env  string
+		want bool
+	}{
+		{"true enables", false, "true", true},
+		{"one enables", false, "1", true},
+		{"mixed case accepted", false, "True", true},
+		{"whitespace-padded value accepted", false, " true ", true},
+		{"false disables a yaml enable", true, "false", false},
+		{"garbage disables rather than guessing", true, "yes-please", false},
+		{"unset leaves yaml alone", true, "", true},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("TAG_CACHE_PARQUET_OPTIMIZATION", tc.env)
+			cfg := NewDefault()
+			cfg.Cache.ParquetOptimization = tc.yaml
+			applyEnvOverrides(cfg)
+			if cfg.Cache.ParquetOptimization != tc.want {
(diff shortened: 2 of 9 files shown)


efirs (dismissed):

ovaistariq:
Thanks both — all three findings were real. Fixed in 99d9b2a.

**Bugbot, "Prefetch skips assembled footer reads" — correct, and it was the important one.** Any single range no longer than one block returns via `serveAssembledRange` before `streamBlockRange` runs, and a parquet reader's first read is the 8-byte trailer. So the trigger sat on a path that the signal it keys on does not take: in the Parseable shapes this targets, it would have essentially never fired.

Worth naming why my tests missed it. `TestPrefetchParquetFooterBlocks_FetchesTheSpannedBlocks` called `prefetchParquetFooterBlocks` directly — it verified the mechanism while *assuming* the wiring. The new `TestServeRangeFromBlockCache_TrailerProbeTriggersFooterPrefetch` drives `serveRangeFromBlockCache` with an actual trailer probe. I checked it fails without the fix (`ranges so far: []` — no upstream request at all) and passes with it.

The assembled path now also attributes prefetch hits for the covering blocks it read from cache, which had the same gap: prefetched footer blocks would be served there and never counted as used, understating precision exactly where the feature works.

**Greptile, "Prefetch hits are double-counted" — correct.** `Get` then `Remove` is not atomic, so two serves racing on the same block could both increment. `expirable.LRU.Remove` reports presence and clears under one lock, so it is now a single atomic test-and-clear. This mattered beyond the counter drift: I had documented "precision cannot exceed 1" as an invariant, and the race falsified it.

**Greptile, "Feature gate suppresses baseline metric" — the finding is right; I fixed the claim rather than the default.** The PR description said the instrumentation "is unconditional", and it is not — `tag_cache_parquet_footer_bytes` only records while the flag is on. That was my error and the description is corrected.

I kept the gate rather than making measurement unconditional, deliberately: the flag is meant to be the single opt-in for parquet-format-aware behavior, and splitting it would add config surface to answer a one-time canary question. The circularity is bounded in practice — enabling also turns on prefetching, but that work is capped at 32 blocks, skipped entirely when metadata fits the tail block, coalesced per object version, and shed rather than queued under populate pressure. So canary-enabling on one node to read the distribution is safe, and that flow is now written down in `docs/metrics.md`. One thing I did confirm: the histogram is observed *before* the "fits in the tail block" early return, so the canary sees the whole population rather than only the objects that got prefetched — otherwise the measurement would have been selection-biased toward the answer it is meant to test.

ovaistariq:
Correct, and this is the sharpest finding on the PR — it defeated the feature on the one path it was built for. Fixed in 9632b4a.

**Verified the mechanism before fixing.** `serveAssembledRange` → `fetchBlocksForAssembly` → `fetchOneBlockForAssembly`, which waits on `serveDone` rather than `cacheDone` — deliberately, per its own doc comment, so a remote assembled serve doesn't block on durability. So on a cold trailer probe the tail block genuinely is not yet readable from cache when the prefetch goroutine runs. `readParquetFooterLength` missed, treated the miss as eviction, and returned silently.

The consequence is worse than a missed optimization. The prefetch would have fired **only on warm re-opens** — the case where every metadata block is already cached and there is nothing to fetch — and never on a first open, where all of them are absent. The feature would have looked harmless and done nothing.

**It also would have corrupted the measurement.** `tag_cache_parquet_footer_bytes` is the metric I argued should decide whether to keep this at all, and it is observed *after* the length is read. Sampling only warm opens biases the footer-size distribution toward objects that were already fully cached. I'd have canaried this on ORD and drawn a conclusion from a skewed sample.

**Fix:** pass the served bytes down and take the trailer from them, falling back to the cache read only when the caller no longer holds them (the streaming path, which has already written them to the client). The trailer is copied out before returning, since `buf` is pooled and the fetch runs detached.

**On the test gap, which you also called correctly.** `TestPrefetchParquetFooterBlocks_FetchesTheSpannedBlocks` seeded the tail block, so it could never have hit this. The new `TestPrefetchParquetFooterBlocks_UsesServedTrailerWhenTailNotYetCached` seeds *nothing* and asserts the two metadata reads still go out. I checked it fails without the change — `prefetched ranges = []`, the exact silent skip — and passes with it.

That is twice on this PR that my tests validated the mechanism while assuming the wiring around it. Both regression tests now start from an empty cache and drive the real path.

`make test` green across all packages, `make lint` clean, `-race` clean on the parquet and block-cache suites.

ovaistariq:
@greptile

ovaistariq:
**Correcting my own rationale on this PR: I had the central premise backwards, and measuring settled it.**

I'd written that metadata fitting inside the tail block was "expected to be the common case", and argued on that basis that the histogram was worth more than the prefetch. That was a guess, and it was wrong.

I sampled 26 objects from `tigris-parseable-prod-ord-1` across two days, reading each trailer directly:

| Object size | Footer | Blocks spanned (1 MiB) |
| --- | --- | --- |
| 394 MB | **4.69 MB** | ~5 |
| 342 MB | **4.31 MB** | ~5 |
| 269 MB | **3.53 MB** | ~4 |
| 244 MB | **3.01 MB** | ~3 |
| 0.2–1.9 MB | 19–52 KB | 0 (fits) |

Footers are a consistent **~1.25% of object size** — the stream schema is 76 KB wide, with per-row-group column statistics on top. **The prefetch fires on 18 of 26 objects**, and on the large ones the tail block covers barely a fifth of the metadata.

There was a second error underneath the first: I'd been comparing footer size against `block_size`. The real condition is `footer + 8 > ContentLength mod block_size` — the **remainder** block, which averages half a block. So the bar is lower than I'd stated on both counts.

**What changes in the PR:** the file-level comment, the `ParquetOptimization` config doc, `docs/metrics.md` and `docs/configuration.md` now carry the measured numbers instead of the guess, and the stale "share of objects above 1 MiB — near zero means turn it off" query is gone (1054956).

**What doesn't change:** the code. The logic already tested the remainder block correctly (`firstBlock >= tailBlock`), and the histogram is still worth keeping — but for **drift detection** rather than for deciding whether to enable. It now has a known-good baseline, so a schema change that moves footer size, or a distribution collapsing below the remainder size, shows up as a signal that the optimization has stopped paying.

The rollout note is simplified accordingly: the footer-size question that would have gated a canary is already answered, so enabling on one ORD node is now about measuring precision (`prefetch_used / prefetched`) and first-open miss rate, not about finding out whether there's anything to do.

PR #177, opened 2026-08-21
feat(cache): warm parquet footers at write time

ovaistariq:
Implements [RFC 0002](docs/rfcs/0002-parquet-footer-warming.md). Adds a second trigger under the existing `cache.parquet_optimization` flag: cache a parquet object's metadata blocks when TAG proxies the **write**, instead of waiting for a reader to miss them.

## Why the read-triggered prefetch isn't enough

The prefetch added in #174 is reactive by construction — the first read of a file is what fires it, so it can only help the second and later reads. For a **sliding-window** reader, the first read is the only one that misses.

A dashboard querying `[now - W, now]` every `R` re-reads the same window, so:

- The window's files are read `W/R` times, and every read after the first finds them cached.
- The window **slides**: each interval, newly written files enter it and old ones leave.

So the only cold footer reads are files written since the last refresh — and a read-triggered prefetch cannot help those, because for them the triggering read *is* the cold read.

This is a common shape wherever ingesters continuously write parquet and a dashboard queries recent data through the same cache. When one TAG cluster proxies both sides, **the write is an exact predictor of the read**.

## How it works

One suffix-range read supplies everything:

```
GET <object>   Range: bytes=-8
```

| From | Gives |
| --- | --- |
| body | the 8-byte trailer → metadata length |
| `Content-Range: bytes X-Y/Z` | `Z`, the object size — no `HEAD` needed |
| `ETag` header | keys the blocks and the meta entry |

The **suffix** form is the point: it needs no prior knowledge of the object's size. That is what lets this run for an object TAG has never seen, without a second round trip and without coupling TAG to any writer's catalog format.

Blocks are then populated through the existing block-mode path, so the RFC 0001 ordering is untouched: **blocks first, tombstone-aware meta last** as the visibility gate.

## Bounds — inherited, not reinvented

- Block-count cap; blocks already cached are skipped.
- Coalesced per object, so a retried `CompleteMultipartUpload` warms once.
- Populate budget taken **non-blocking** — warming sheds rather than queues when real reads contend.
- Fires after the client's write response is committed, so it adds no write latency.
- Anonymous writes skipped (a public write implies nothing about read access).
- A `.parquet` suffix is a hint, not a guarantee — a bad trailer aborts silently.

## Why not `warm_on_write`

It already exists and caches **whole objects**. Footers are a small fraction of object size — on a wide-schema deployment, measured around 1.25% — so this moves roughly 1% of the bytes for the benefit that actually matters here.

## Measurement is the decision rule

No new metric families. Reuses `tag_cache_block_prefetched_total{trigger="write_warm"}` against `tag_cache_block_prefetch_used_total`, so precision is directly comparable to `parquet_footer`'s.

If `write_warm` precision comes in materially lower, readers aren't following writers in that deployment and the trigger should be turned off. That's the stated rule in the RFC — the metric decides, not the argument.

`tag_cache_parquet_footer_bytes` is how a deployment tells whether its schema makes any of this worthwhile: a wide schema puts several MB of metadata on a large object, a narrow one keeps it inside the tail block where there is nothing to prefetch.

## Review fixes

Both bots caught a real bug and I'd documented the opposite of what the code did:

- **Non-206 drained the full body.** The deferred `io.Copy(io.Discard, ...)` ran on the `200` early-return too, so a server ignoring the range pulled the entire object through the warm path — exactly the cost the suffix range exists to avoid, and the opposite of what the comment claimed. Now the `200` path closes **without** draining, and the `206` drain is bounded.
- **Range interval unvalidated.** Only the total was checked, so bytes from a different interval could have been parsed as a trailer and used to derive block indices. Now the interval must be exactly `[total-8, total-1]`, with a regression test that feeds trailer-shaped bytes from the wrong offset.
- Reused the existing `parseContentRange` from `service.go` rather than adding a second parser.

## Testing

`parquetFooterBlocks` (spans metadata through the tail, footer-inside-tail, pathological-footer cap), `readParquetTrailerFromUpstream` (resolves size/ETag/footer from one suffix range; rejects a `200`, a missing ETag, a wrong interval, and a non-parquet trailer), and trigger gating (disabled, non-parquet key, retried write coalesces).

## Rollout

Off by default. Where `cache.parquet_optimization` is already enabled for #174, merging and deploying turns this on too — worth watching `write_warm` precision against `parquet_footer` for the first hour. Reverting is a config change, not a rollback.

---

> [!NOTE]
> **Medium Risk**
> Touches cache populate and successful write paths (PUT, copy, multipart complete) and issues extra ranged GETs with write credentials. Gated, off by default, and does not change the serve path.
> 
> **Overview**
> Adds a **write-time** parquet footer warm (RFC 0002) under the existing `cache.parquet_optimization` flag, so metadata blocks are cached when TAG proxies the upload instead of only after a reader misses.
> 
> After a successful `PutObject`, copy, or `CompleteMultipartUpload`, a detached job issues one suffix-range GET (`bytes=-8`) to learn size, ETag, and footer length, then populates the spanning blocks through the existing block-mode path. Work runs after the client response, coalesces per key, skips anonymous writes, and does not drain a `200` that ignored the range.
> 
> Prefetch counters now distinguish `write_warm` vs `parquet_footer`. Block-mode meta stamping is shared via `finalizeBlockModeMeta` so warming can attribute blocks before the RFC 0001 visibility gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7fed8cc39af396ee4a691d4784c5dc836ab3282. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -181,14 +181,18 @@ cache:
   # default (1 MiB). Override with TAG_CACHE_BLOCK_SIZE env var.
   block_size: 1048576
 
-  # Parquet-aware footer prefetching (opt-in). A parquet reader must read the file's metadata
+  # Parquet-aware footer caching (opt-in). Drives two triggers: a read that reaches an
+  # object's trailer prefetches the rest of its metadata, and a parquet object written
+  # through TAG has its metadata warmed before the first read (RFC 0002).
+  # A parquet reader must read the file's metadata
   # before any data, and that metadata sits at the end of the object. Block caching already
   # covers metadata that fits in the tail block; this fetches the earlier blocks when it does
   # not, using the length the file itself declares rather than a guess. Costs one speculative
   # fetch per spanned block, shed under populate pressure like any other prefetch. Watch
-  # tag_cache_parquet_footer_bytes to confirm it still applies to your data:
-  # measured on production parseable data, footers run ~1.25% of object size, so a
-  # 300 MB object carries ~3.5 MB of metadata -- several blocks at a 1 MiB block_size.
+  # tag_cache_parquet_footer_bytes to see whether it applies to your data: footer
+  # size scales with row groups and columns, so a wide schema can put several MB of
+  # metadata on a few-hundred-MB object -- several blocks at a 1 MiB block_size --
+  # while a narrow schema keeps it inside the tail block and this does nothing.
   # Override with TAG_CACHE_PARQUET_OPTIMIZATION env var.
   parquet_optimization: false
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -375,10 +375,17 @@ means the readers are not following writes closely, and that trigger is not payi
 
 **Type:** Histogram
 
-Size of the parquet metadata region, read from the trailer of objects served while
+Size of the parquet metadata region, read from the trailer of objects while
 `cache.parquet_optimization` is on. Recorded for **every** parquet object whose trailer is
 read, including ones that are not prefetched, so the distribution describes the whole
 population rather than just the prefetched tail of it.
+
+Note the population depends on which triggers are active. The read trigger observes objects
+as they are read; the write trigger observes them as they are written. With both on, an object
+written and later read through the same node is observed twice, and write-heavy deployments
+will weight the distribution toward recently written objects. Read it as a distribution of
+*observations*, not of distinct objects — that is what makes it useful for drift detection and
+what to keep in mind before comparing across deployments.
 
 The prefetch does work whenever `footer + 8` exceeds the **remainder** block
 (`ContentLength mod block_size`, averaging half a block) — not merely when the footer exceeds
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -361,7 +361,15 @@ precision cannot exceed 1. It is also deliberately an undercount — attribution
 bounded, TTL-expiring set, so a block served long after it was prefetched goes unattributed.
 Read the ratio as a floor.
 
-The only `trigger` today is `parquet_footer`, from `cache.parquet_optimization`.
+Both triggers come from `cache.parquet_optimization`:
+
+| `trigger` | Fires when |
+| --- | --- |
+| `parquet_footer` | A read reached a parquet object's trailer — reactive, helps the second and later reads |
+| `write_warm` | A parquet object was written through TAG — proactive, removes the first read's miss (RFC 0002) |
+
+Compare the two directly. `write_warm` precision materially below `parquet_footer` precision
+means the readers are not following writes closely, and that trigger is not paying for itself.
 
 #### tag_cache_parquet_footer_bytes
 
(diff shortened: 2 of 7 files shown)


ovaistariq:
Both bots caught the same real bug, and it's the worst kind: **my comment documented the opposite of what the code did.**

```go
defer func() {
    _, _ = io.Copy(io.Discard, resp.Body)   // ← runs on EVERY return
    _ = resp.Body.Close()
}()
if resp.StatusCode != http.StatusPartialContent {
    // "do not drain a hundreds-of-MB body just to find a trailer"
    return nil, 0, false                     // ← drains a hundreds-of-MB body
}
```

The drain was there for connection reuse, which is the right instinct for an 8-byte response and exactly wrong for a whole object. A server ignoring the suffix range would have pulled the entire object through the warm path on every write — defeating the design and, under write load, spiking bandwidth. Fixed: the `200` path closes **without** draining, and the `206` drain is bounded rather than read-to-EOF.

**Greptile's P2 on the unvalidated interval is also right**, and matters more than it first looks: block indices are derived from those bounds. Only the total was checked, so a server answering a different interval could have had its bytes parsed as a trailer. Now the interval must be exactly `[total-8, total-1]`, with a regression test that returns trailer-shaped bytes from offset 0 and asserts rejection. While fixing it I found `parseContentRange` already existed in `service.go` with the right semantics, so I deleted my duplicate and reused it.

**On the P2 about Info-level logging for new features — I'm not taking this one**, and I want to explain rather than just decline.

That rule was learned from `tigris-os#3115`, but this repo made the opposite call in [#150](https://github.com/tigrisdata/tag/pull/150) for exactly this shape of code. The finding there was that Info-logging per-item errors "for new-feature visibility" floods the log during an upstream outage — precisely when the logs are most needed — and that TAG's cache errors should be Debug **and propagated**, with a metric carrying the visibility instead.

This path is per-write. An upstream outage would make every write log an error at Info. The visibility this rule wants already exists and is better: `tag_cache_block_prefetched_total{trigger="write_warm"}` flatlining is an alertable signal that warming has stopped, without a log line per failure.

Happy to be overruled if the team's view has changed since #150, but I don't think a per-write path should be the place we reverse it.

ovaistariq:
Both correct. Fixed in 03aef50.

**"Warm path stores incomplete object meta" (High) — this was a production panic, and the sharpest finding on the PR.**

I hand-rolled the entry with the three fields the block arithmetic needs:

```go
&cache.CachedObjectMeta{ETag: etag, BlockSize: ..., ContentLength: ...}
```

But that meta is the **visibility gate** for the entry, so it is what a later serve reads back. `service.go:634` calls `w.WriteHeader(meta.StatusCode)` on a HEAD cache hit — with `StatusCode: 0` that panics net/http (`invalid WriteHeader code 0`). It also dropped `Bucket`, `Key`, `Content-Type`, `Last-Modified` and user metadata that clients expect on a served object.

The root mistake was building a second, divergent way to construct an entry the range path already builds correctly. `buildBlockMeta` runs `MetaFromHTTPHeaders` over the 206's headers and stamps `StatusCode: 200` (the object status, not the partial one). Now used here too, so there is one constructor rather than two.

The new test asserts the meta carries a `WriteHeader`-valid status and its own identity. I confirmed it fails against the previous code — `StatusCode = 0, which panics net/http at WriteHeader` — rather than assuming it passes.

**"Write warm dedup ignores object version" (Medium) — right that the comment lied**, though I'd put the impact slightly lower than stated.

The key is `pqw:bucket/key` while the comment claimed one warm per object *version*. It cannot be keyed by version: the ETag is only learned from the trailer read that happens *inside* the warm. So a second write landing mid-warm is skipped, as you describe.

Where I'd differ is "the new version is never warmed" — the read-triggered path from #174 still warms it on its first read. So the cost is **one cold read**, not a permanent gap. That is an acceptable trade for keeping the retry-coalescing that the key exists for, and the comment now states the behaviour and the fallback instead of claiming something stronger.

Worth noting both of these came from the same habit: reaching for a hand-written struct or key instead of the primitive the read path already uses. The meta fix removes the duplicate constructor entirely.

ovaistariq:
⚠️ **Superseded — please do not merge this.** Everything here is already on `main`, and merging it now would *revert* #178.

I branched `proto/meta-on-write` (#178) off this branch, so when #178 was **squash-merged** as f38c06f it carried every commit from here along with it. `main` already has the footer warming, RFC 0002, and the four review-pass fixes that landed on top.

Verified rather than assumed:

- `d7fed8c` (this branch's tip) is an ancestor of the branch #178 was cut from.
- `git diff main...this-branch` is **721 deletions against 49 insertions**, and the insertions are all *reversions* — restoring the pre-review versions of `prefetchParquetFooterBlocks`, the un-cooled-down dedup key, and the untriggered attribution, while deleting `meta_on_write.go` and its measurement entirely.

So the merge button here is a revert button. Closing rather than leaving it open.

The work is not lost — it is on `main` in a better state than this branch, since #178 also carried the seven-finding review pass, the attribution atomicity fix, and the race test that actually fails when the mutex is removed.

Worth noting the process trap for next time: branching a follow-up PR off an unmerged PR makes the second one silently contain the first. That is convenient while iterating and misleading afterwards — the parent PR looks open and mergeable while being a revert.

ovaistariq:
Handled through #178

PR #179, opened 2026-08-21
refactor(cache): one footer implementation for both triggers, plus a user guide

ovaistariq:
Two things: collapse the duplicated footer logic into a single implementation, and add a user-facing guide for the parquet optimization.

## The duplication was the defect source

The read trigger (`parquet_footer`) and the write trigger (`write_warm`) each carried their own copy of *"work out which blocks the metadata spans, fetch the absent ones, count what was fetched"*. Across #174 and #178 that duplication is where most of the findings came from — the two copies drifted in exactly the bookkeeping that decides whether the feature is judged to work, and each fix had to be applied twice or was applied once.

Both now funnel through `ensureParquetFooterBlocks`. They differ only in **how they learn the footer length** — a served or cached trailer on read, a suffix-range GET on write — which is the one thing that genuinely differs between them.

Net: `block_cache_parquet.go` loses the second copy of the block-range maths, the presence filter, the counting, and the histogram observation.

## Two behaviour changes fall out, both improvements

**Concurrent fetches.** The shared path uses `fetchBlocksToCache`, so footer blocks are fetched in parallel rather than one at a time. Three tests asserted the serial ordering; order was never part of the contract, so they compare as sets now.

**The read trigger gains stale-meta invalidation it did not have.** `fetchBlocksToCache` invalidates an entry whose ETag no longer matches upstream, with the comment:

> Invalidate HERE — every block fetch flows through this point — so no caller can forget it.

The read path's hand-rolled per-block loop had forgotten it. A prefetch hitting an ETag mismatch left the stale entry in place to fail every later read until TTL.

## One difference was real, and is now named rather than erased

On a **read**, the serve that fired the prefetch already owns the tail block and may still be writing it. Re-fetching it would duplicate that transfer on every cold open — and presence alone cannot distinguish "absent" from "being written right now".

So the caller states it (`tailServedByCaller`) rather than the cache guessing. The old code expressed this same rule twice, differently, and got it right only by accident: the read path hard-coded "skip the tail", the write path hard-coded "include it".

## Verification

Beyond the suite passing, I re-checked that the consolidated path still fails on the two original wiring bugs from #174 — removing the assembled-path trigger, and removing the served-trailer fallback, each still breaks its regression test. A refactor that quietly disarmed those guards would be worse than the duplication it removed.

## docs/parquet-optimization.md

A user-facing guide, linked from the README and cross-linked from `configuration.md` and `metrics.md`. It covers what the optimization does and why parquet needs it, **how to decide whether it applies to your data before enabling it**, how to turn it on, and how to read the precision metrics afterwards.

It is deliberately explicit that the answer may be "leave it off": the comparison is footer size against the **remainder** block (`ContentLength mod block_size`), not against `block_size`, and a narrow schema keeps the footer inside the tail block where there is nothing to prefetch. The histogram is presented as the way to find out, with the expected precision range so an operator can tell a working deployment from a wasteful one.

---

> [!NOTE]
> **Medium Risk**
> Touches parquet footer populate (cache correctness, stale-meta invalidation) and removes a shipped metric plus serve-path attribution. Behavior of which blocks are fetched is intended to stay equivalent, but operators relying on `prefetch_used` lose that signal.
> 
> **Overview**
> Collapses read-triggered (`parquet_footer`) and write-triggered (`write_warm`) parquet footer warming into one path, `ensureParquetFooterBlocks`, so both triggers share block-range math, presence filtering, counting, and fetches.
> 
> The shared path uses `fetchBlocksToCache`, so footer blocks fetch concurrently and the read trigger now invalidates stale meta on ETag mismatch (the old per-block loop skipped that). `parquetFooterBlocks` takes `tailServedByCaller` so the read path still skips the tail block the serve already owns, while the 32-block cap applies after that drop.
> 
> Removes prefetch-precision bookkeeping: `tag_cache_block_prefetch_used_total`, the expirable LRU, and per-block lookups on the serve path. Prefetch is now a volume signal; operators are told to judge payoff from the existing block hit ratio. Adds `docs/parquet-optimization.md` and updates metrics/RFC/config docs to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8faacdbb95e082ad43faba0e16236bb333ca4f63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ For production, TAG ships manifests, Compose files, and guides under [`deploy/`]
 - **Kubernetes** — StatefulSet, HPA, and services in [`deploy/kubernetes/`](deploy/kubernetes/); high availability, scaling, and probes in [docs/deploy.md](docs/deploy.md).
 - **Docker** — single-node and cluster Compose in [`deploy/docker/`](deploy/docker/); see [docs/docker.md](docs/docker.md).
 - **TLS/HTTPS** — see [docs/tls.md](docs/tls.md).
+- **Parquet workloads** — caching parquet metadata ahead of the reader that needs it; see [docs/parquet-optimization.md](docs/parquet-optimization.md).
 - **Benchmarks** — see [docs/benchmarks.md](docs/benchmarks.md).
 
 ## Architecture
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -191,7 +191,7 @@ cache:
   # Override with TAG_CACHE_META_ON_WRITE env var.
   meta_on_write: false
 
-  # Parquet-aware footer caching (opt-in). Drives two triggers: a read that reaches an
+  # Parquet-aware footer caching (opt-in). Full guide: docs/parquet-optimization.md Drives two triggers: a read that reaches an
   # object's trailer prefetches the rest of its metadata, and a parquet object written
   # through TAG has its metadata warmed before the first read (RFC 0002).
   # A parquet reader must read the file's metadata
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -362,7 +362,8 @@ precision cannot exceed 1. It is also deliberately an undercount — attribution
 bounded, TTL-expiring set, so a block served long after it was prefetched goes unattributed.
 Read the ratio as a floor.
 
-Both triggers come from `cache.parquet_optimization`:
+Both triggers come from `cache.parquet_optimization` (see
+[docs/parquet-optimization.md](parquet-optimization.md) for the full guide):
 
 | `trigger` | Fires when |
 | --- | --- |
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -342,36 +342,27 @@ sum(rate(tag_cache_block_prefetch_windows_total{outcome!="full"}[5m])) /
 sum(rate(tag_cache_block_prefetch_windows_total[5m]))
 ```
 
-#### tag_cache_block_prefetched_total / tag_cache_block_prefetch_used_total
+#### tag_cache_block_prefetched_total
 
-**Type:** Counter (both labeled by `trigger`)
+**Type:** Counter (labeled by `trigger`)
 
-Speculative block fetches, and how many of them were later served from cache. Together they
-give prefetch precision — the fraction of speculative work that paid for itself:
+Blocks fetched speculatively rather than to satisfy a request in flight. This is a
+**volume** signal — how much speculative work each trigger is doing — not a precision
+signal.
 
-```promql
-# Prefetch precision, per trigger. A low ratio means that trigger is evicting more
-# than it earns: turn it off rather than tuning it. Both counters carry the label,
-# so the triggers are compared directly rather than pooled.
-sum by (trigger) (rate(tag_cache_block_prefetch_used_total[30m])) /
-sum by (trigger) (rate(tag_cache_block_prefetched_total[30m]))
-```
+Judge the benefit from the block hit ratio (`tag_cache_block_hits_total` /
+`tag_cache_block_misses_total`), which measures the outcome directly. Rising hit ratio
+with modest prefetch volume is what you want; volume with no movement in hit ratio
+means the speculation is not landing where reads go.
 
-Attribution counts blocks, not reads: a prefetched block served many times counts once, so
-precision cannot exceed 1. It is also deliberately an undercount — attribution is tracked in a
-bounded, TTL-expiring set, so a block served long after it was prefetched goes unattributed.
-Read the ratio as a floor.
+There is deliberately no matching `_used_total`. Attributing each serve back to a
+prefetch required a per-block lookup on the serve path, which runs thousands of times
+per second, to re-derive a ratio that is ~98% by construction — a reader that has read
+a parquet trailer cannot do anything except read the metadata region next.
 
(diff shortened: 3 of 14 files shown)


ovaistariq:
All three correct. Fixed in ccb8ace.

**"Partial fetches lose attribution" (Greptile P1 / Bugbot Medium) — the one that mattered, and it is a property the consolidation quietly dropped.**

`fetchBlocksToCache` can cache some blocks and shed or fail on others; I only recorded attribution when it returned `nil`. Those cached blocks are real work, and skipping them made it invisible to the ratio this feature is judged on — **permanently**, because a later attempt cannot credit them either: they no longer look absent.

The old per-block read loop counted as each block landed. Consolidating onto a batch primitive lost that without my noticing, which is a fair criticism of the refactor rather than of the original design. The batch does not report per block, so the blocks are now re-probed afterwards and every one that landed is counted, whatever the aggregate result was.

**"Read prefetch cap off-by-one" (Low) — correct, and the asymmetry is the tell.** The cap was applied to a span that still included the tail, which the read trigger then dropped, so the read path fetched at most 31 blocks where the write path fetched 32 from the same configured limit. The tail is now accounted for before the cap, so the cap bounds what a caller actually fetches. The new test fails with `read trigger got 31 blocks` against the previous arrangement — I checked rather than assumed.

**"Aligned objects use full tails" (P2) — right, and it was in the section most likely to be acted on.** The guide gave the tail as `ContentLength mod block_size`, which is 0 when the length is an exact multiple — implying every non-empty footer on such an object needs prefetching. The tail is a **full block** in that case. That section exists specifically so an operator can decide *before* enabling anything, so a formula that overstates the benefit is the worst place to be loose.

Corrected to the real expression, with the aligned case called out explicitly. Worth noting the **code was already right** — `parquetFooterBlocks` uses integer division on block indices and never computes a modulus — so this was documentation drifting from behaviour rather than a bug.

ovaistariq:
@greptile

ovaistariq:
Correct, and the consequence is worse than a duplicate count. Fixed in b123dc7.

The two triggers use different work keys (`pq:` and `pqw:`), so they do not coalesce with **each other** — only with themselves. A read prefetch and a write warm can therefore overlap on one object, both await the same coalesced *block* fetch, and both find the block present afterwards. One physical transfer, credited twice.

**Why that is not merely cosmetic:** the numerator can only be claimed once, since `notePrefetchHit` clears the entry. So a transfer counted twice reads as **precision 0.5 for work that succeeded completely** — in the metric documented as the rule for deciding whether a trigger stays enabled. It would have argued for turning off a trigger that was working.

Claiming and counting are now one atomic, idempotent step: only the call that claims a block counts it. The increment moved *inside* `notePrefetchedBlock`, which is the substance of the fix rather than a tidy-up — the counter and the attribution set are the denominator and numerator of one ratio, and keeping them in separate places is exactly what let them disagree. **This is the third defect in this series where those two drifted apart**, so the structure now prevents it instead of each call site having to remember.

Worth being clear that this is **pre-existing, not introduced by the consolidation**: the old read and write paths counted independently too, so the same double-credit was possible. Consolidating them into one function is what made it visible.

The test overlaps both triggers on one block across 100 objects and asserts prefetched equals the number of transfers. Against the previous unconditional increment it reports `200 prefetched blocks for 100 transfers` and `precision = 100/200` — verified, not assumed.

ovaistariq:
@greptile

ovaistariq:
Both correct, and they share one root cause worth fixing properly rather than patching either symptom. Fixed in a4292bd.

**Cache presence proves a block arrived; it never proves who brought it.** Inferring "this trigger fetched it" from a post-fetch `BlockExists` is wrong in both directions, which is why the two findings look opposed but are the same bug:

- **Over-counts** (Greptile): a footer prefetch that joins a fetch already in flight — the other trigger's, or an ordinary demand read's — credits itself for a transfer it did not cause.
- **Under-counts** (Bugbot): the claim-once gate I added to suppress that over-count then suppresses a *legitimate* re-warm, where the block was evicted and genuinely re-fetched while its attribution entry was still live.

Both inflate precision, against a metric this repo documents as a **floor**. Patching either one in isolation would have moved the error to the other.

**The signal already existed and was simply never surfaced.** `beginBlockFetch` distinguishes the caller that creates a fetch from one that joins an in-flight state — it just returned only the state. It now reports ownership, `fetchOneBlock` and `fetchBlocksToCache` pass it through, and the prefetch counts exactly the transfers it owned. Four call sites that do not care ignore it with `_`. Ownership also makes the claim-once gate unnecessary, so it is gone and a re-warm counts again.

**The test needed replacing, not updating.** The previous one simulated coalescing by calling `notePrefetchedBlock` twice — which the new contract forbids — so it was asserting a *convention* rather than a behaviour. It now drives the real path: two callers overlap on one block behind a gated upstream fetch, and it asserts one transfer produces one owner. Against presence-based attribution it reports `2 callers claimed ownership of 1 transfer`, and it also asserts the block was fetched upstream exactly once.

**One residual, stated rather than hidden:** an owner whose fetch short-circuits because the block appeared between the presence check and the fetch is still counted. Callers filter on presence first, so that window is small — but it is not zero, and I would rather write that down than imply the count is exact.

ovaistariq:
@greptile

PR #181, opened 2026-08-21
fix(cache): invalidate a block-less entry on the amplify bail after all

ovaistariq:
Greptile flagged a **P1 on the release PR** (#180), against code already on `main` from #178. It is correct, and it reverses a decision I made two PRs ago.

## The two findings point opposite ways

- **#178, Bugbot:** a full-object GET over a metadata-only entry bails, invalidates, and destroys the entry meta-on-write just established — making the feature useless against full GETs. I added an exemption: don't invalidate when the entry holds **no** blocks.
- **Now, Greptile:** that exemption leaves stale **metadata** behind. If the object was deleted or replaced out of band, later HEADs answer `200` with stale headers until the TTL (24h by default).

Both describe real consequences of the same code. What settles it is that only one of them is a correctness problem.

## Why my original reasoning was wrong

I argued that an entry with no cached blocks has nothing to serve stale. That is true of its *bytes* and false of its *metadata*.

The deeper issue: **the bail decides from a probe and never contacts upstream.** It cannot know whether the object still exists. `errBlockUpstreamGone` handles exactly this case — a block fetch that 404s invalidates the entry — but the bail never fetches, so it never learns. I was choosing between two guesses and picked the one that keeps a cache entry.

Losing the entry costs **one discovery round trip on the next read**. Keeping a stale one **answers wrongly**. That is not a close call.

## The change

The exemption is gone, and with it `errNoRequestedBlocksCached`, which existed only to carry it — `ensureBlocksCached` returns the plain amplify error again. Net effect on `block_cache.go` is a deletion.

The test is inverted: it now asserts the entry does **not** survive an unvalidated full GET, and its failure message says why (`a deleted object would answer HEAD 200 until TTL`).

## What this costs meta-on-write

A full-object GET of a written-but-unread object discards its entry, and the next read re-pays the discovery round trip. Ranged reads — the parquet access pattern this targets — are unaffected, since they never take the full-object bail.

If that turns out to matter, the right fix is to *learn* rather than guess: validate upstream before deciding, or invalidate from the fall-through's own response status. Both are real changes rather than a heuristic, and neither belongs in a fix for a stale-read bug.

**#180 will need re-cutting** (or re-merging `main`) to pick this up.

---

> [!NOTE]
> **Medium Risk**
> Changes cache invalidation for full-object GETs over incomplete block entries, which can drop live meta-on-write entries and affect subsequent HEAD/GET correctness until re-discovery.
> 
> **Overview**
> Always invalidate a block-mode entry when a full-object GET bails on mostly-missing blocks, including **metadata-only** entries that previously were exempt.
> 
> The amplify bail never talks to upstream, so it cannot tell if the object was deleted or overwritten. Keeping a block-less entry avoided discarding meta-on-write work, but it also let later HEADs return 200 with stale headers until TTL. Losing the entry costs one discovery round trip; keeping a stale one is wrong.
> 
> Removes `errNoRequestedBlocksCached` (it existed only for that exemption). `invalidateStaleBlockMeta` now records delete success/failure instead of ignoring cache delete errors. The test now asserts a deleted origin object does not leave a surviving entry after an unvalidated full GET.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6392830825cc7bed10fdac22405bbdcd36b8d077. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxy/block_cache.go
+++ b/proxy/block_cache.go
@@ -46,14 +46,6 @@ var errBlockUpstreamGone = errors.New("block upstream gone")
 // as "fall through to the miss path", not a real error.
 var errBlockAssemblyWouldAmplify = errors.New("block assembly would amplify")
 
-// errNoRequestedBlocksCached is the amplify bail for a request whose covering blocks
-// are ALL absent. The predicate is per-request, not per-entry: only a full-object
-// serve spans every block, so only there does it mean the entry itself holds nothing.
-// It wraps errBlockAssemblyWouldAmplify, so callers that only care about "fall
-// through" match it unchanged, while the full-object path uses the distinction to skip
-// an invalidation that would protect nothing (see serveCompleteFromBlocks).
-var errNoRequestedBlocksCached = fmt.Errorf("%w: none of the requested blocks are cached", errBlockAssemblyWouldAmplify)
-
 // maxInlineFetchesPerServe bounds how many absent blocks one COMMITTED block serve recovers via
 // individual aligned upstream fetches. BlocksComplete is a hint: blocks and meta evict/expire
 // independently, so a surviving meta over mass-evicted blocks would otherwise turn one full GET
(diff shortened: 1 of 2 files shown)


efirs (dismissed):

ovaistariq:
Both correct. Fixed in the latest commit.

**The flaky test (Bugbot + Greptile, same issue) — right, and the diagnosis is exact.** I primed a live object and polled for the entry to disappear, but the miss path re-populates after the bail invalidates. When that repopulation landed before the first poll, correct behaviour failed the assertion purely on timing. Worse, absence was the wrong property to assert at all: what matters is that the **old unvalidated entry** does not answer later reads, and a fresh upstream-validated entry appearing in its place is fine.

Deleting the object upstream fixes both problems at once. There is nothing to re-populate, so absence is stable — and it models the actual hazard rather than a proxy for it: an entry outliving the object it describes. I confirmed it still fails when the invalidation is skipped, reporting `entry for a deleted object survived an unvalidated full GET`.

**"Invalidation failures remain silent" (P1) — also right, and it was a plain omission.** `invalidateStaleBlockMeta` called `s.cache.Delete(...)` and discarded the return entirely: no check, no log, no metric. That delete is precisely what stops a stale entry answering later reads, so a silent failure leaves the hazard the call exists to prevent — and a delete metric that only ever reports success would hide it.

`invalidateObject` already handles its own failures this way, with a comment stating the reasoning ("a false-green delete metric would hide the very read-after-write hazard the invalidation exists to prevent"). This now matches it rather than having a second, quieter convention.

ovaistariq:
@greptile

